PR #7, opened 2022-12-07
2 add build scripts

K20shores:
Builds a single static library for musica core and tuv-x. This will make adding other components to CAM easy. We could include json-fortran if we'd like to keep our library self-contained. Looks like cmake has support for that, but I'll only do that if we think it's a good idea.


CMakeLists.txt
```diff
@@ -0,0 +1,167 @@
...
+
+################################################################################
+# Projet wide setup options
+include(CMakeDependentOption)
+
+option( ENABLE_TUVX "Builds TUV-x, a photolysis calculator library" ON )
+option( ENABLE_TESTS "Builds tests that ensures each enabled MUSICA component can be used" ON )
+option( ENABLE_MPI "Enable MPI parallel support" OFF)
+option( ENABLE_OPENMP "Enable OpemMP support" OFF )
+# cmake_dependent_option( ENABLE_OPENMP "Enable OpemMP support" OFF "ENABLE_MPI" OFF)
```
mattldawson, line 29: could probably just remove this

mattldawson (approved):
looks great!

PR #19, opened 2023-05-24
17 put musica in an environment module

K20shores:
This stubs out an infrastructure to create a C interface to musica components. Further, the beginnings of a CCPP module for micm is started. Lastly, I wrote a dockerfile that compiles all of MUSICA components and the C interface with GCC and then compiles the CCPP parts of MUSICA with the intel fortran compiler. A test is run to make sure the intel fortran compiled CCPP test can call the GCC compiled musica library.


Dockerfile.ccpp
```diff
@@ -0,0 +1,67 @@
+# on macOS this image requires rosetta 2 support for virtualization
```
K20shores, line 1: @WardF this is the docker file that builds the intel compilers. You'd only need the first 35 lines

boulderdaze (approved):

mattldawson (approved):

PR #30, opened 2023-07-24
Fortran wrapper

boulderdaze:
Completed

- Removed the `ccpp` contents
- Added `musica-fortran`
- `musica` is built and installed by gcc. `musica-fortran` is built by intel compilers and linked to `musica` 

Notes
- `netcdf` is also needed by `musica-fortran` to build


src/fortran_interface/micm/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+target_sources(musica
```
K20shores, line 1: make this be musica-fortran

src/fortran_interface/micm/interface.F90
```diff
@@ -0,0 +1,19 @@
...
+!      subroutine solver(state, state_size, time_step) bind(c)
+!         import :: c_ptr, c_double, c_int64_t
+!         real(c_double), dimension(*) :: state
+!         integer(c_int64_t), value :: state_size
+!         integer(c_int64_t), value :: time_step
+!      end subroutine
+
+   end interface
+
+end module micm
```
K20shores, line 19: This is correct, but to compile you'll need to add this
```suggestion
module micm
    use iso_c_binding

    implicit none
   
   interface

!      type(c_funptr) function get_solver(filepath) bind(c)
!         import :: c_char, c_funptr
!         character(len=1, kind=c_char), dimension(*), intent(in) :: filepath
!      end function get_solver

!      subroutine solver(state, state_size, time_step) bind(c)
!         import :: c_ptr, c_double, c_int64_t
!         real(c_double), dimension(*) :: state
!         integer(c_int64_t), value :: state_size
!         integer(c_int64_t), value :: time_step
!      end subroutine

   end interface

end module micm
```

CMakeLists.txt
```diff
@@ -117,6 +115,37 @@ target_include_directories(musica
 
 add_subdirectory(src)
 
+#
+# jiwon 7/21 - in progress
+#
+####################
+# MUSICA - fortran
+add_library(musica-fortran STATIC src/fortran_interface/micm/interface.F90)
```
K20shores, line 123:
```suggestion
add_library(musica-fortran)
```

The file itself will be added with the suggested change to the CMakeLists file in srd/fortran_interface

CMakeLists.txt
```diff
@@ -117,6 +115,37 @@ target_include_directories(musica
...
+    $<BUILD_INTERFACE:${MUSICA_MOD_DIR}>
+    $<INSTALL_INTERFACE:${INSTALL_MOD_DIR}>
+)
+
+# add the sources to musica
+target_sources(musica
+  PRIVATE
+    $<TARGET_OBJECTS:musica-fortran>
+)
+
```
K20shores, line 148: The goal is to use the fortran part of musica separately from musica. So, the fortran objects should not be a part of the `musica` cmake target.
```suggestion
```

CMakeLists.txt
```diff
@@ -117,6 +115,37 @@ target_include_directories(musica
...
+set_target_properties(musica-fortran PROPERTIES
+  ARCHIVE_OUTPUT_DIRECTORY ${MUSICA_LIB_DIR}
+  Fortran_MODULE_DIRECTORY ${MUSICA_MOD_DIR}
+)
+
+target_link_libraries(musica-fortran 
+  PUBLIC
+    PkgConfig::netcdff
+    ${JSON_LIB}
+)
```
K20shores, line 135: I don't think we'll need this since we aren't linking to json or netcdff for this little target
```suggestion
```

CMakeLists.txt
```diff
@@ -117,6 +115,37 @@ target_include_directories(musica
...
+
+target_include_directories(musica-fortran
+  PUBLIC  
+    $<BUILD_INTERFACE:${MUSICA_MOD_DIR}>
+    $<INSTALL_INTERFACE:${INSTALL_MOD_DIR}>
+)
```
K20shores, line 141: We may not end up needing this, not sure yet.

CMakeLists.txt
```diff
@@ -117,6 +115,37 @@ target_include_directories(musica
...
+# jiwon 7/21 - in progress
+#
+####################
+# MUSICA - fortran
+add_library(musica-fortran STATIC src/fortran_interface/micm/interface.F90)
+add_library(musica::musica-fortran ALIAS musica-fortran)
+
+set_target_properties(musica-fortran PROPERTIES
+  ARCHIVE_OUTPUT_DIRECTORY ${MUSICA_LIB_DIR}
+  Fortran_MODULE_DIRECTORY ${MUSICA_MOD_DIR}
```
K20shores, line 128: It's possible we'll want these to be new cmake variables, but we will figure that out when we make an example project

K20shores (changes requested):

CMakeLists.txt
boulderdaze: Comments Added for this

CMakeLists.txt
boulderdaze: Comments Added for this

CMakeLists.txt
```diff
@@ -1,224 +1,224 @@
-################################################################################
```
K20shores, line 1: What happened to using cmake variables to selectively build musica or musica fortran? Did we decide that's not a good idea?

K20shores: Preserving the usability of musica with fetch content will not work without a top level cmake file

Dockerfile.fortran
```diff
@@ -0,0 +1,109 @@
...
+RUN cd musica/musica-fortran \
+    && cmake -S . \
+             -Bbuild \
+             -D ENABLE_TESTS=ON \
+    && cd build \
+    && make
+
+######
+######
+# TODO: jiwon 8/4 - We need these but it takes too long time to build so I will skip for now
```
K20shores, line 56: We can copy the micm file for intel https://github.com/NCAR/micm/blob/main/Dockerfile.intel. I didn't realize when I originally wrote this that intel published docker images

src/CMakeLists.txt
```diff
@@ -7,4 +7,5 @@ target_sources(musica
     ${CMAKE_BINARY_DIR}/version.c
 )
 
-add_subdirectory(cxx_interface)
\ No newline at end of file
+add_subdirectory(cxx_interface)
+add_subdirectory(fortran_interface)
```
K20shores, line 11: is there a fortran interface here anymore?

CMakeLists.txt
K20shores: If it's too messy to separate them cleanly maybe we should just create yet another repository

K20shores (changes requested):

K20shores (approved):

Dockerfile.fortran
```diff
@@ -0,0 +1,70 @@
...
+    && cd build-intel \
+    && cmake -D SKIP_DOC_GEN:BOOL=TRUE .. \
+    && make install
+
+# Set json-fortran variable to build MUSICA-Fortran using intel
+ENV JSON_FORTRAN_HOME="/usr/local/jsonfortran-intel-8.2.0"
+
+RUN cd musica/musica-fortran/test \
+    && mkdir build && cd build \
+    && cmake .. \
```
mattldawson, line 67: Does this not need the `USE_MUSICA_FORTRAN=ON`? Or maybe I'm not following the logic correctly

mattldawson (approved):
looks great! just one minor question, but I think I just was just not following the logic correctly

Dockerfile.fortran
K20shores: It's an entirely new cmake project. Internally it uses fetch content to pull musica and in that cmake file it sets use_musica_fortran=on

PR #35, opened 2023-09-25
Added creating solver interface

boulderdaze:
Completed
- Created fortran interface for micm
-  Current version only supports 1 grid cell


musica-fortran/src/micm/MICM_mod.F90
```diff
@@ -0,0 +1,69 @@
+module libmicm
```
K20shores, line 1:
```suggestion
module micm
```

please rename to micm. Our fortran style is to call the module the component name and any types <module_name>_t

musica-fortran/src/micm/MICM_mod.F90
```diff
@@ -0,0 +1,69 @@
...
+    use iso_c_binding
+
+    private
+    public :: micm
```
K20shores, line 5:
```suggestion
    public :: micm_t
```

musica-fortran/src/micm/MICM_mod.F90
```diff
@@ -0,0 +1,69 @@
...
+        implicit none
+        type(micm) :: this
+        call delete_micm_c(this%ptr)
+    end subroutine
+    
+    subroutine delete_micm_polymorph(this)
+        implicit none
+        class(micm) :: this
+        call delete_micm_c(this%ptr)
+    end subroutine
```
K20shores, line 50: Is this needed?

boulderdaze: My understanding is that yes, we need this. I also had the same question and without it, it doesn't work. It needs to declare micm as `class` instead of `type` to be able to have access to its members.  I found [this answer](https://stackoverflow.com/questions/53421609/polymorphism-in-fortran) useful to understand.

musica-fortran/src/micm/CMakeLists.txt
```diff
@@ -1,4 +1,6 @@
 target_sources(musica-fortran
   PUBLIC
-  interface.F90
```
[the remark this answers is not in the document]
boulderdaze: Addressed

musica-fortran/src/micm/MICM_mod.F90
boulderdaze: Fixed

musica-fortran/src/micm/MICM_mod.F90
```diff
@@ -0,0 +1,69 @@
+module libmicm
+    use iso_c_binding
+
+    private
+    public :: micm
+
+    include "MICM_C_def.F90"
+
+    type micm
```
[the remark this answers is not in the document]
boulderdaze: Fixed

musica-fortran/src/micm/MICM_mod.F90
```diff
@@ -0,0 +1,69 @@
...
+        private
+        type(c_ptr) :: ptr
+    contains
+        procedure :: delete => delete_micm_polymorph
+        procedure :: create_solver => micm_create_solver
+        procedure :: solve => micm_solve
+    end type
+
+    interface micm
```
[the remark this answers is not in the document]
boulderdaze: Fixed

musica-fortran/src/micm/MICM_mod.F90
boulderdaze: I updated it to `class` and removed `type` version.

musica-fortran/src/micm/MICM_mod.F90
boulderdaze: Fixed

musica-fortran/src/micm/micm_mod.F90
```diff
@@ -0,0 +1,63 @@
...
+    private
+    public :: micm_t
+
+    include "micm_c_def.F90"
+
+    type micm_t
+        private
+        type(c_ptr) :: ptr
+    contains
+        procedure :: delete => delete_micm
```
K20shores, line 13: @mattldawson is this a time when `finalize` could/should be used?

musica-fortran/src/micm/micm_mod.F90
```diff
@@ -0,0 +1,63 @@
...
+        procedure :: solve => micm_solve
+    end type
+
+    interface micm_t
+        procedure create_micm
+    end interface
+
+contains
+    function create_micm(config_path)
+        implicit none
```
K20shores, line 24: Please place `implicit none` at the top of this on the line before `private` and remove `implicit none` in all of these functions

K20shores (changes requested):

musica-fortran/src/micm/micm_mod.F90
boulderdaze: Yes, it's done now.

musica/src/micm/micm.cpp
```diff
@@ -0,0 +1,74 @@
...
+    else
+    {
+        std::cout << "   * [C++] Failed creating solver" << std::endl;
+        success = 0; 
+    }
+
+    return success;
+}
+
+void MICM::delete_solver() const
```
K20shores, line 46: Please make this a destructor so that in `delete_micm` you can just call `delete micm` and not have to care about calling this function

K20shores (changes requested):

musica-fortran/src/micm/micm_mod.F90
mattldawson: yeah, I think so - if this function is always called right before the object is deleted, it could be:
```suggestion
        final :: delete_micm
```

mattldawson: the function itself will have to be modified to take a `type(micm_t)` instead of a `class(micm_t)` with `intent(inout)`, but other than that should be mostly the same (I think). This `final` function gets called automatically when a `micm_t` object goes out of scope

musica/include/micm/micm.hpp
```diff
@@ -0,0 +1,39 @@
...
+    MICM(const std::string& config_path);
+    ~MICM();
+
+    // TODO(jiwon): can return type indicate error?
+    int create_solver();
+    void delete_solver() const;
+    void solve(double temperature, double pressure, double time_step, double*& concentrations, size_t num_concentrations);
+
+private:
+    static constexpr size_t NUM_GRID_CELLS = 1; // TODO(jiwon)
```
mattldawson, line 24: we can set up a separate issue for supporting multiple grid cells

mattldawson (approved):
looks great! I don't see anything other than what Kyle commented on

musica-fortran/src/micm/micm_mod.F90
boulderdaze: Would it be okay to leave the way it is then? It needs to be `class(micm_t)` for `micm_t` to have access to its members.

mattldawson: I think you should still have access to the data members, you just would need to change it to:
```
    subroutine delete_micm(this)
        type(micm_t), intent(inout) :: this
        call delete_micm_c(this%ptr)
    end subroutine
```
I think the only difference between `class(foo_t)` and `type(foo_t)` is that `class(foo_t)` will accept types that extend `foo_t`, but `type(foo_t)` won't, and that type-bound procedures have to take the polymorphic `class` argument, but final routines need to take the concrete `type` argument.

boulderdaze: Ah, thank you for correcting me. I fixed it

musica/src/micm/micm.cpp
boulderdaze: Yeah, I updated the destructor.

K20shores (approved):

PR #88, opened 2024-04-12
Add custom error handling option

### LINKED ISSUE #38: Add error handling for MUSICA interface
Add error handling for MUSICA interface

- For any functions in the MUSICA library that can catch exceptions thrown by MICM, TUV-x, etc., allow host applications to use their own logic to handle the errors

**Acceptance Criteria**
- In the Atmospheric_Physics wrapper of MUSICA, any errors coming from MICM functions can be written out to the atmosphere log file

**Ideas**
Either:
- return an error code and message from any function that catches exceptions
or:
- Accept a function pointer as an argument to any function that catches exceptions. The function passed in should accept two arguments - an error code and a message. The function will be called if/when an exception is caught.
or:
- Host applications can register a function that will be called whenever an exception is caught. MUSICA could have a default function that fails with a message to stderr whenever an exception is caught.

Example error-handling function:
```
void HandleError(int error_code, int error_message_length, const char* error_message);
```

Then, a host application could register this function with MUSICA something like:
```
RegisterErrorHandler(HandleError);
```

mattldawson:
Adds an option for users to provide a custom error handling function through the C API that will be called for any exceptions thrown by the C++ libraries.

closes #38 

I'm not sure if this is the best design. I think it will work for the CAM-SIMA use case where they just want to provide a single function that gets called for any error. But, I'm not sure if a global function pointer for errors is the best solution. I don't mind changing this approach entirely if anyone has a better solution.


K20shores:
The only thing I really dislike is random numbers we have to provide as the `code` argument. I can't think of a better way right now

fortran/micm_core.F90
```diff
@@ -120,16 +124,16 @@ function constructor(config_path, errcode)  result( this )
       end do
       c_config_path(n+1) = c_null_char
 
-      this%ptr = create_micm_c(c_config_path, errcode)
-
-      if (errcode /= 0) then
```
K20shores, line 125: A difference here is that we are taking control of the error flow in C++. Whereas what exists now allows for cam-sima to check the result of the error code and control the error flow. I think this makes it confusing to use our c api

fortran/test/fetch_content_integration/test_micm_fort_api.F90
```diff
@@ -34,7 +35,7 @@ program test_micm_fort_api
 
 
   write(*,*) "[test micm fort api] Creating MICM solver..."
-  micm => micm_t(config_path, errcode)
+  micm => micm_t(config_path)
```
K20shores, line 38: There's no recourse here for when creating micm fails. The error function is called, I think, but the program still thinks that micm is valid.

src/micm/micm.cpp
```diff
@@ -1,180 +1,207 @@
...
-        return nullptr;
-    }
+MICM *create_micm(const char *config_path) {
+  MICM *micm = new MICM();
+  try {
+    micm->create_solver(std::string(config_path));
+    return micm;
+  } catch (const std::exception &e) {
+    delete micm;
+    Error(909039518, std::string("Error creating MICM solver: " + std::string(e.what())).c_str());
```
K20shores, line 23: Since these codes are really meant to make where the error happened identifiable, maybe this is another time where a macro is useful and we use `__LINE__` and `__FILE__` to do this instead?

mattldawson:
> The only thing I really dislike is random numbers we have to provide as the `code` argument. I can't think of a better way right now

Yeah, I agree. After more discussion and checking with the CAM-SIMA developers, I think I might redo this PR to just have the MUSICA C API functions always return an error code and message by catching any exceptions thrown in MICM. I will also explore options that don't require us to create random numbers for error codes.

mattldawson:
I think this is ready for another look now. All the API functions return an error object that has a code, category, and message. The combination of the code and category can be used to identify specific errors, and there are some helper functions to do this. Let me know what you think!

K20shores (approved):
I like it!

src/util.cpp
```diff
@@ -12,7 +12,62 @@ String ToString(char* value)
...
+}
+
+Error NoError()
+{
+    Error error;
+    error.code_ = 0;
+    error.category_ = ToConstString("");
+    error.message_ = ToConstString("Success");
+    return error;
+}
```
mwaxmonsky, line 35:
```suggestion
Error NoError()
{
    return ToError(ToConstString(""), 0, ToConstString("Success"));
}
```

src/util.cpp
```diff
@@ -12,7 +12,62 @@ String ToString(char* value)
...
+
+Error ToError(const std::system_error& e)
+{
+    Error error;
+    error.code_ = e.code().value();
+    error.category_ = ToConstString(e.code().category().name());
+    error.message_ = ToConstString(e.what());
+    return error;
+}
```
mwaxmonsky, line 58:
```suggestion
Error ToError(const std::system_error& e)
{
    return ToError(ToConstString(e.code().category().name()),
                   e.code().value(),
                   ToConstString(e.what()));
```

mwaxmonsky (approved):
Looks great @mattldawson!

Had two suggestions to use base functions in the error handler that might be worth while but nothing that should hold up the PR.

src/util.cpp
mattldawson: done!

src/util.cpp
mattldawson: done!

boulderdaze (approved):
Looks good!

PR #106, opened 2024-05-11
Update build scripts

mattldawson:
Modifies build scripts to allow for optional inclusion of MICM and/or TUV-x, and cross-compiler linking of installed MUSICA-C library with MUSICA-Fortran wrapper.

The NVidia fetch content integration test still fails, and it may be a compiler problem. I'll open a separate issue for this.


K20shores (approved):

boulderdaze (approved):
Looks good!

PR #113, opened 2024-05-15
Add code to allow for verbose fetchcontent.

WardF:
By default, `FETCHCONTENT_QUIET` is set to `TRUE`.  With good bandwidth, the underlying git actions invoked when running `cmake` to configure `musica` are transparent.  When on very slow internet, however, the `cmake [srcdir]` step can take nearly 10 minutes due to the size of the dependency downloads.

Setting `FETCHCONTENT_QUIET=FALSE` will insert status lines into the cmake-generated output, but by default it does not show the `git` status re: download.  This PR results in more verbose output when `-DFETCHCONTENT_QUIET=FALSE` is passed to `cmake` at configure time.

### DIFF
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -1,5 +1,5 @@
 include(FetchContent)
-
+#SET(FETCHCONTENT_QUIET FALSE)
 ################################################################################
 # NetCDF library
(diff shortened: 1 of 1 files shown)


mattldawson (approved):

WardF:
Superceded by #124, closing as redundant.

PR #132, opened 2024-05-29
MUSICA Tutorial Chapter 1

dwfncar:
Tutorial Chapter 1, linking a fortran program to the MUSICA library.,
Added get_micm_version function to the fortran interface.


CMakeLists.txt
```diff
@@ -70,6 +70,7 @@ elseif(${CMAKE_Fortran_COMPILER_ID} MATCHES "GNU")
 elseif(${CMAKE_Fortran_COMPILER_ID} MATCHES "PGI")
   list(APPEND musica_compile_definitions MUSICA_USING_PGI)
 endif()
+set(CMAKE_Fortran_FLAGS "-fPIE")
```
K20shores, line 73: if we need to set this, then please set it with `[target_compile_options](https://cmake.org/cmake/help/latest/command/target_compile_options.html)` on the appropriate target (musica fortran?). This prevents us from polluting downstream builds with options that they may not want.

docs/source/getting_started/overview.rst
```diff
@@ -0,0 +1,20 @@
...
+and a box model in a single modular framework.
+The model infrastructure will be open source,
+flexible and computationally efficient in order
+to facilitate community co-development and use for scientific and operational purposes. 
+
+At the heart of MUSICA is the standalone Model Independent Chemistry Model (MICM)
+that satisfies the requirements of the Common Community Physics Package (CCPP)
+and that can be connected to any CCPP compliant atmosphere model.
+MUSICA and MICM will have a flexible design to handle a variety of gas phase and aerosol schemes
+and associated chemical modules such as deposition or photolysis.
```
K20shores, line 20:
```suggestion
At the heart of MUSICA is the standalone Model Independent Chemistry Model (MICM), which is a gas-phase kinetic solver. MICM is made available by the MUSICA wrapper which satisfies the requirements of the Common Community Physics Package (CCPP)
and that can be connected to any CCPP compliant atmosphere model.
MUSICA and MICM will have a flexible design to handle a variety of gas phase and aerosol schemes
and associated chemical modules such as deposition or photolysis.
```

K20shores (changes requested):
I like it!

CMakeLists.txt
dwfncar: I've taken this out for now ... I don't understand why I needed it on one Linux OS (RedHat) with GCC 13.2 but not on another (Ubuntu) with GCC 11.4.

K20shores (approved):

docs/source/user_guide/fortran_c.rst
```diff
@@ -0,0 +1,78 @@
+##############
+Model Inteface
+##############
+
+Fortran C Interface Example
+---------------------------
```
mattldawson, line 6: I don't see where this ends up in the documentation.

docs/source/tutorial/chapter1.rst
```diff
@@ -0,0 +1,51 @@
...
+.. code-block:: f90
+
+  program demo
+      use musica_util, only: string_t
+      use musica_micm, only: get_micm_version
+      implicit none
+      type(string_t) :: micm_version
+      micm_version = get_micm_version()
+      print *, "MICM version ", micm_version%get_char_array()
+  end program demo
```
mattldawson, line 21: I would suggest that we move all the tutorial code snippets to separate source code files that get compiled and run as tests. This way we can ensure that if we break a part of the tutorial with updates to the MUSICA source code, a test will fail.

We do something similar in MICM:

https://github.com/NCAR/micm/blob/ee5c03e904d9dc3d60619dc9dbce06232bd32852/docs/source/user_guide/rate_constant_tutorial.rst?plain=1#L46-L49

mattldawson (changes requested):
Looks great! Just one question and one suggestion.

boulderdaze (approved):
I really like the overview!

docs/source/user_guide/fortran_c.rst
dwfncar: The section is a place holder for now ...

docs/source/tutorial/chapter1.rst
dwfncar: Yes, let me do that ...

docs/source/user_guide/fortran_c.rst
mattldawson: ah, that makes sense. thanks!

fortran/test/fetch_content_integration/test_get_micm_version.F90
```diff
@@ -0,0 +1,8 @@
+program test_get_micm_version
+    use musica_util, only: string_t
+    use musica_micm, only: get_micm_version
+    implicit none
+    type(string_t) :: micm_version
+    micm_version = get_micm_version()
+    print *, "MICM version ", micm_version%get_char_array()
+end program test_get_micm_version
```
mattldawson, line 8: Looks great! I would just now pull the content of this test directly into the tutorial so they don't get out-of-sync. Similar to this:
https://github.com/NCAR/micm/blob/ee5c03e904d9dc3d60619dc9dbce06232bd32852/docs/source/user_guide/rate_constant_tutorial.rst?plain=1#L46-L49

Also, I would move this to a `test/tutorial/` folder or something similar

mattldawson (changes requested):
Looking better! Just one last comment

dwfncar:
Now using literal include for test_get_micm_version.F90 in Chapter 1.
Creating of test/tutorial subdir and CMakeList files to move to Chapter 0 issue.

mattldawson (approved):
looks great!

PR #140, opened 2024-06-05
Make tuvx grids available in musica fortran and musica c++

### LINKED ISSUE #108: NVidia tests failing
After updating the Dockerfile for the NVidia build to use the gcc-built MUSICA C library with an NVidia-built MUSICA Fortran wrapper, the NVidia tests are failing.

### LINKED ISSUE #120: Allow TUV-x grid data to be updated through the API
Add the ability for host applications to update grid data at runtime.

**Acceptance Criteria**
- Users can set the edge and mid-point values for any grid in the TUV-x configuration (heights, wavelengths(read-only))

**Ideas**
- Add `grid_map_t` and `grid_t` types to the `tuvx_interface` module, something like:
```
type :: grid_map_t
  type(c_ptr) :: ptr_ ! Pointer to internal tuvx_core%grid_warehouse_
end type

type :: grid_t
  type(c_ptr) :: ptr_ ! Pointer to specific grid in tuvx_core%grid_warehouse_
end type
```
- Add `internal_get_grids_tuvx` function to `module tuvx_interface`:
```

    function internal_get_grids_tuvx(tuvx) result(grids)
      use iso_c_binding, only: c_ptr, c_f_pointer
      type(c_ptr), intent(in) :: tuvx
      type(grid_map_t) :: grids

      type(core_t), pointer :: core

      call c_f_pointer(tuvx, core)
      grids%ptr_ = c_loc( core%grid_warehouse_ )

    end function internal_get_grids_tuvx

```
- do something similar for `grid_t` that looks up grids by name and units
- add `grid_t::update_edges()` and `grid_t::update_mid_points()`
- a user could do something like this:
```
   tuvx => tuvx_t( config_path )
   grids = tuvx%create_grids( )
   grid = grids%get( "height", "m" )
   call grid%update_edges( new_edge_data_array )
   call grid%update_mid_points( new_mid_point_data_array )
```

image

K20shores:
Closes #120, #146, also relies on https://github.com/NCAR/tuv-x/pull/90 and https://github.com/NCAR/tuv-x/pull/96
Apparently also closes #108

- Creates external fortran facing types for grid, grid_maps
- Allows external fortran to get grids and grid maps
- Passes all information along to the internal tuv types

What I don't have, as of submission, is a test that checks for data being set properly because I don't know if we want to do what I did (which is copy the memory). If we stick with that, we will need a getter which will return a copy(?) to the external facing forcing and then we can check.

An alternative is to pass only the pointer around so that the memory is shared completely between fortran->C++->fortran. That seems like a bit problem waiting to happen though. 

Should I create the getter or share the memory?


fortran/musica_micm.mod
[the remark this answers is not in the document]
K20shores: yes, not even sure why this is here

cmake/dependencies.cmake
```diff
@@ -74,7 +74,7 @@ if (MUSICA_ENABLE_TUVX AND MUSICA_BUILD_C_CXX_INTERFACE)
   set(TUVX_INSTALL_INCLUDE_DIR ${MUSICA_INSTALL_INCLUDE_DIR} CACHE STRING "" FORCE)
 
   set_git_default(TUVX_GIT_REPOSITORY https://github.com/NCAR/tuv-x.git)
-  set_git_default(TUVX_GIT_TAG 9900942d2ccfc9cc03516e406bda6f77a4da7d66)
+  set_git_default(TUVX_GIT_TAG add_grid_warehouse)
```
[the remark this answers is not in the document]
K20shores: It is until https://github.com/NCAR/tuv-x/pull/90 is merged

src/tuvx/tuvx.cpp
```diff
@@ -14,58 +14,194 @@
...
+      *error = ToError(e);
     }
-    catch (const std::exception &e)
+    catch (...)
     {
-      parsing_status = 2;
-      std::cerr << e.what() << std::endl;
-      DeleteString(&config_path_str);
+      *error = Error{ 1, CreateString(MUSICA_ERROR_CATEGORY), CreateString("Failed to create tuvx instance") };
+    }
```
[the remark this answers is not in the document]
K20shores: well on that note all of our errors are leaking memory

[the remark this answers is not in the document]
boulderdaze: Could you elaborate what you mean? I think we should try to  prevent the memory leak as mush as possible and this leak caused by `String` is fairly obvious.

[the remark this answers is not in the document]
K20shores: `CreateString` allocates memory for a `char*` array. Typically we pass this over to fortran. I am not sure if we are deallocating the memory each time. It's entirely possible that we are with all of the micm stuff and I failed to replicate it with tuv. I also started another branch to check the string handling and fix it if it's wrong. If I find anything, I'll make a PR from that branch into this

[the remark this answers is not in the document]
boulderdaze: Ah okay, sounds good, thank you.

boulderdaze (approved):

src/tuvx/tuvx.cpp
```diff
@@ -14,58 +14,194 @@
...
+    tuvx->Create(std::string(config_path), error);
+    if (!IsSuccess(*error))
     {
-      std::cerr << "File doesn't exist: " << config_path << std::endl;
-      *error_code = 2;
+      delete tuvx;
       return nullptr;
     }
+    return tuvx;
+  }
```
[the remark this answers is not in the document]
K20shores: I put that check into `TUVX::Create`. It seems more appropriate there since these functions are really just pass through things for C. I think we should keep them small and light, but I can move it here if you think it's better here

cmake/dependencies.cmake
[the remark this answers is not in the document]
K20shores: Fixed

src/tuvx/tuvx.cpp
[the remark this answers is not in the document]
K20shores: So we don't leak memory when we are on the successful path. Everytime we pass one of these Strings to fortran, fortran calls another function `delete_string_c` that deletes the allocated memory after fortran copies it to some local memory.

However, if fortran runs into an error before it calls delete_string_c, then we will deleter memory.


I changed it to this
```
      std::unique_ptr<String, decltype(&DeleteString)> config_path_str(new String, &DeleteString);
      config_path_str->size_ = config_path.size();
      config_path_str->value_ = new char[config_path.size() + 1];
      std::strcpy(config_path_str->value_, config_path.c_str());
      tuvx_ = InternalCreateTuvx(*config_path_str, &parsing_status);
```

and we are guarnteed to always delete the string when it goes out of scope in case of an error. It won't be a double free because the `DeleteString` function checks if the `value_` of the string is still allocated. It's not thread safe though, but I don't think we need this function to be thread safe

build.log
```diff
@@ -0,0 +1,227 @@
+#0 building with "desktop-linux" instance using docker driver
```
mattldawson, line 1: looks like this file could be removed

fortran/test/unit/tuvx.F90
```diff
@@ -41,6 +41,15 @@ subroutine test_tuvx( )
...
 
+!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+  !> Test TUV-x API
+  subroutine test_tuvx_api()
+
+    
+
+  end subroutine test_tuvx_api
+
```
mattldawson, line 52: is this for a future test?

fortran/tuvx.F90
```diff
@@ -1,64 +1,337 @@
...
+   public :: tuvx_t, grid_map_t, grid_t
+   private
+
+   !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+   interface
+      function create_tuvx_c(config_path, error) bind(C, name="CreateTuvx")
+         use musica_util, only: error_t_c
+         import c_ptr, c_int, c_char
+         character(kind=c_char), intent(in) :: config_path(*)
```
mattldawson, line 21:
```suggestion
         character(len=1, kind=c_char), intent(in) :: config_path(*)
```

src/tuvx/interface.F90
```diff
@@ -46,4 +70,132 @@ end subroutine internal_delete_tuvx
...
+
+!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+    function interal_get_grid(grid_map, c_grid_name, c_grid_name_length, c_grid_units, c_grid_units_length, error_code) &
+        result(grid_ptr) bind(C, name="InternalGetGrid")
+      use iso_c_binding, only: c_ptr, c_f_pointer, c_int, c_char, c_size_t
+    
+      ! arguments
+      type(c_ptr), intent(in), value                   :: grid_map
+      character(kind=c_char), dimension(*), intent(in) :: c_grid_name
```
mattldawson, line 102:
```suggestion
      character(len=1, kind=c_char), dimension(*), intent(in) :: c_grid_name
```

src/tuvx/interface.F90
```diff
@@ -46,4 +70,132 @@ end subroutine internal_delete_tuvx
...
+      integer(kind=c_size_t), value                    :: c_grid_name_length
+      character(kind=c_char), dimension(*), intent(in) :: c_grid_units
```
mattldawson, line 104:
```suggestion
      character(len=1, kind=c_char), dimension(*), intent(in) :: c_grid_units
```

mattldawson (approved):
looks great!

fortran/test/fetch_content_integration/test_tuvx_api.F90
```diff
@@ -1,51 +1,100 @@
...
+
+    call grid%set_edges( edges, error )
+    ASSERT( error%is_success() )
+
+    call grid%set_midpoints( midpoints, error )
+    ASSERT( error%is_success() )
+    
+    ! profile => profiles%get( "O3", "mol m-3" )
+
+    ! call tuvx%solve( grids, profiles, radiators, photo_rates )
```
boulderdaze, line 93: Can the commented code be removed?

fortran/test/fetch_content_integration/test_tuvx_api.F90
```diff
@@ -1,51 +1,100 @@
...
+  subroutine test_tuvx_api_invalid_config()
+    type(tuvx_t), pointer :: tuvx
+    type(error_t)         :: error
+    character(len=256)     :: config_path
 
     config_path = "invalid_config"
 
-    tuvx => tuvx_t(config_path, errcode)
+    tuvx => tuvx_t(config_path, error)
+    ! ASSERT( .not. error%is_success() )
```
boulderdaze, line 52: Can the commented code be removed?

fortran/test/fetch_content_integration/test_tuvx_api.F90
```diff
@@ -1,51 +1,100 @@
...
-  end subroutine test_tuvx_fort_api
+  end subroutine test_tuvx_api
 
   ! Invalid tuvx solver creation test
-  subroutine test_tuvx_fort_api_invalid()
-    character(len=7) :: config_path
+  subroutine test_tuvx_api_invalid_config()
+    type(tuvx_t), pointer :: tuvx
+    type(error_t)         :: error
+    character(len=256)     :: config_path
```
boulderdaze, line 47: Maybe spacing could be uniform?

include/musica/tuvx.hpp
```diff
@@ -41,14 +97,24 @@ namespace musica
...
    public:
+    TUVX();
+
     /// @brief Create an instance ove tuvx from a configuration file
     /// @param config_path Path to configuration file or directory containing configuration file
-    /// @return 0 on success, else on failure in parsing configuration file
-    int Create(const std::string &config_path);
+    /// @param error Error struct to indicate success or failure
+    /// @return 0 on success, 1 on failure in parsing configuration file
+    void Create(const char* config_path, Error *error);
```
boulderdaze, line 106: There seems to have a typo and an old comment that should be removed. 
```C++  
  /// @brief Create an instance of tuvx from a configuration file
  /// @param config_path Path to configuration file or directory containing configuration file
```

valgrind.supp
```diff
@@ -7,25 +7,17 @@
...
    ...
-   fun:internal_create_tuvx
+   fun:InternalCreateTuvx
    ...
 }
 {
   <AssertFailureLeaksString>
   Memcheck:Leak
-  match-leak-kinds: definite
+  match-leak-kinds: possible
```
boulderdaze, line 19: This is interesting, could you explain what the change of `possible` from `definite` do?

boulderdaze (changes requested):
Looks really good! Thank you for fixing some other issues on the way

fortran/test/unit/tuvx.F90
K20shores: oh I forgot about this. All of the tests ended up in the fetch content integration tests. I'll remove this

valgrind.supp
K20shores: when I ran valgrind, the new message changed from definite to possible. Not entirely sure why

include/musica/tuvx.hpp
boulderdaze: Could you remove the `line 105`?
```c++
/// @return 0 on success, 1 on failure in parsing configuration file
```

boulderdaze (approved):

WardF (approved):
Looks good!

PR #160, opened 2024-06-21
Add air density to MICM solve function as input

### LINKED ISSUE #157: Air density should be an input to `MICM::Solve()`
Air density [mol m-3] should be an input similar to temperature and pressure for the `MICM::Solve()` function

**Acceptance Criteria**
- Air density is always passed by users to MICM's solve function through the C, Fortran, and Python interfaces

mattldawson:
Add air density as an input to the solve function for MICM

closes #157


boulderdaze (approved):
Looks good!

PR #165, opened 2024-06-25
158 create an action to test the build and packaging of the python target

### LINKED ISSUE #158: Create an action to test the build and packaging of the python target
Add an action that uses PIP to build the Python target.

**Acceptance Criteria**
- A test runs on each commit to main that tests the building of the Python target using PIP

**Ideas**
- Follow logic here [pybind11 project](https://github.com/pybind/cmake_example) and [here](https://github.com/pybind/cmake_example/blob/7a94877f581a14de4de1a096fb053a55fc2a66bf/.github/workflows/pip.yml)

K20shores:
Closes #158 

Additionally, the cmake version is parsed by the docs and the pyproject.toml files so that we only need to update the version number in one place and documentation and python releases will use the same version.


docker/Dockerfile.python
```diff
@@ -1,23 +1,45 @@
...
+               " > requirements.txt 
+
+RUN python -c "import toml; \
+               file = 'musica/pyproject.toml'; \
+               data = toml.load(file); \
+               deps = data.get('project', {}).get('optional-dependencies', {}).get('test', []); \
+               [print(dep) for dep in deps]; \
+               " >> requirements.txt
+
+RUN pip install -r requirements.txt
```
boulderdaze, line 42: Is this valid? Looks like `requirements.txt` is deleted and replaced with a call to install `numpy`

boulderdaze (commented):
Looks good!

docker/Dockerfile.python
K20shores: the `>>` appends and doesn't overwrite

boulderdaze (approved):

mattldawson (approved):

PR #172, opened 2024-07-03
Vector-ordered Rosenbrock Solver

### LINKED ISSUE #55: Vector-ordered Rosenbrock Solver
Allow the use of the MICM Rosenbrock solver with vector-ordered matrices through the MUSICA C library and Fortran wrapper

**Acceptance Criteria**
- At compile time, users can specify a vector dimension (the `L` template parameter on vectorizable classes).
- Users can choose to use a Vector-ordered Rosenbrock solver and it will have the `L` value they chose at compile time.

**Ideas**
- Since the vector dimension must be known at compile time and C/Fortran don't have C++ template-equivalents, the current `MICM_DEFAULT_VECTOR_MATRIX_SIZE` can be used to set the `L` value .
- Add a GH test to the MUSICA repo to set the vector dimension at compile time and run the MICM tests

boulderdaze:
Closes #55
- Update `musica` build options
- Doesn't support multi grid cells and vector matrix size greater than 1. (the way assigning the concentrations in Solve function) 
- I decided to submit this with the limitation described above because this task currently blocks implementing Backward Euler as well as a separate issue was created to support multiple grid cells.    

- These comments are not valid anymore but left for record to describe where the bug found. 
Added compilation flag for the number of grid cells: `MUSICA_SET_NUM_GRID_CELLS` 
 For vector-ordered matrix solver, when number of grid cell is less than `L`  value, for example,  `num_grid_cell=1` , `L=4` , `num_species=5` , it fails during solving because the [Copy()](https://github.com/NCAR/micm/blob/main/include/micm/util/vector_matrix.hpp#L324-L329) call throws an error indicating that the dimension is not the same; copying other vector to this vector `(5 /= 20)`. Since `L` is decided at compile time, checks should be included in the build script


CMakeLists.txt
```diff
@@ -30,6 +30,15 @@ option(MUSICA_BUILD_DOCS "Build the documentation" OFF)
 option(MUSICA_ENABLE_MICM "Enable MICM" ON)
 option(MUSICA_ENABLE_TUVX "Enable TUV-x" ON)
 
+set(MUSICA_SET_MICM_NUM_GRID_CELLS "1" CACHE STRING "Set the number of grid cells for MICM Solver")
```
mattldawson, line 33: The number of grid cells is something that won't be know at compile time for the MUSICA library, if a host model is linking against an installed version of MUSICA.

Can this be passed as an argument to the MICM constructor?

fortran/test/fetch_content_integration/test_micm_api.F90
```diff
@@ -42,21 +43,22 @@ subroutine test_api()
     real(c_double), parameter     :: GAS_CONSTANT = 8.31446261815324_c_double ! J mol-1 K-1
     integer                       :: i
     
+    config_path = "configs/chapman"
+    solver_type = 1
```
mattldawson, line 47: I'm a little hesitant to just have an integer determine the solver type. We will eventually want to have many of the solver options be configurable. What about having a struct for solver options, that can be used to specify the vector vs standard matrix, number of grid cells, ODE algorithm, etc.? maybe we could discuss this as a group to come up with a long-term strategy? We could always create a separate issue for this too

include/musica/micm.hpp
```diff
@@ -162,20 +202,126 @@ namespace musica
...
+    using RosenbrockStandard = micm::Solver<RosenbrockStandardType, micm::State<DenseMatrixStandard, SparseMatrixStandard>>;
+    std::unique_ptr<RosenbrockStandard> rosenbrock_standard_;
 
+   private:
     std::unique_ptr<micm::SolverParameters> solver_parameters_;
   };
 
+  template<class T>
+  inline void MICM::Solve(
+      std::unique_ptr<T> &solver,
```
mattldawson, line 230: could this just be:
```
auto &solver
```
to avoid having a template parameter?

include/musica/micm.hpp
```diff
@@ -162,20 +202,126 @@ namespace musica
...
+    }
+    catch (const std::system_error &e)
+    {
+      DeleteError(error);
+      *error = ToError(e);
+    }
+  }
+
+  template<class T>
+  inline std::map<std::string, std::size_t> MICM::GetSpeciesOrdering(std::unique_ptr<T> &solver, Error *error)
```
mattldawson, line 290: `auto &solver` here too?

include/musica/micm.hpp
```diff
@@ -162,20 +202,126 @@ namespace musica
...
+    catch (const std::system_error &e)
+    {
+      DeleteError(error);
+      *error = ToError(e);
+      return std::map<std::string, std::size_t>();
+    }
+  }
+
+  template<class T>
+  std::map<std::string, std::size_t> MICM::GetUserDefinedReactionRatesOrdering(std::unique_ptr<T> &solver, Error *error)
```
mattldawson, line 308: `auto &solver` here too?

src/test/unit/micm/micm_c_api.cpp
```diff
@@ -179,8 +192,8 @@ TEST_F(MicmCApiTest, GetUserDefinedReactionRatesOrdering)
   DeleteMappings(reaction_rates_ordering, array_size);
 }
 
-// Test case for solving the MICM instance
-TEST_F(MicmCApiTest, SolveMicmInstance)
+// Test case for solving system using vector-ordered Rosenbrock solver
+TEST_F(MicmCApiTest, SolveUsingVectorOrderedRosenbrock)
```
mattldawson, line 196: should we also have a test with multiple grid cells?

mattldawson (changes requested):
looks good! just a couple questions and suggestions. the only important one is that that the number of grid cells won't be known at compile time

cmake/test_util.cmake
```diff
@@ -43,7 +43,7 @@ function(create_standard_test_cxx)
   add_executable(test_${TEST_NAME} ${TEST_SOURCES})
   target_link_libraries(test_${TEST_NAME} PUBLIC musica::musica GTest::gtest_main)
   if(MUSICA_ENABLE_OPENMP)
-    target_link_libraries(test_${TEST_NAME} PUBLIC OpenMP::OpenMP_CXX OpenMP::OpenMP_Fortran)
+    target_link_libraries(test_${TEST_NAME} PUBLIC OpenMP::OpenMP_CXX)
   endif()
```
boulderdaze, line 47: Hi @mattldawson and @K20shores, what do you think about this change? Matt and I briefly discussed about this. The issue is that `OpenMP::OpenMP_Fortran` is not found when `test_micm_c_api` is built. I removed it because this line belongs to `function(create_standard_test_cxx)` and I assumed that it shouldn't need `OpenMp_Fortran` for c++ tests. 
Matt mentioned we could specify fortran as one of the languages in the top-level `CMakeLists.txt` and update `find_package` in `cmake/dependencies.cmake`
```
project(musica-distribution VERSION 0.7.0 LANGUAGES C CXX Fortran)
find_package(OpenMP REQUIRED COMPONENTS C CXX Fortran)
```
I think that makes sense although I think it is useful to separate the fortran stuff from `MUSICA-C` as it is since we eventually pursue fortran-free. I'd like to hear what you all think.

src/test/unit/micm/micm_c_api.cpp
boulderdaze: Yes, I think so but this PR doesn't support multiple grid cells yet. Do you think your new task of supporting multiple grid cell can add those test?

mattldawson: yes, I agree

CMakeLists.txt
boulderdaze: That's a good point. I made it to be passed as an argument to the MICM constructor.

include/musica/micm.hpp
boulderdaze: Yes, done!

include/musica/micm.hpp
boulderdaze: When I replaced the template parameter with `auto`, it mostly passed the tests except for the following four. (`error: symbol not found in flat namespace`)
```
Mac / c_cxx (g++-11, gcc-11)
Windows / mingw (x64) 
Mac / c_cxx (g++-12, gcc-12)
Mac / c_cxx (g++-13, gcc-13)
```
I guess for some compilers the type of `solver` is not deducible. I found this link is useful [Why can't you use auto as a template type argument](https://stackoverflow.com/questions/18701279/why-cant-you-use-auto-as-a-template-type-argument-e-g-stdarrayauto-5) but still quite a mystery to me why it works for others.

include/musica/micm.hpp
boulderdaze: Same case with `MICM::GetSpeciesOrdering`. It doesn't pass all the test when `auto&` used.

fortran/test/fetch_content_integration/test_micm_api.F90
boulderdaze: I agree. I was uncomfortable with using `int` to represent the solver type but only did in favor of fortran. I updated it using enum type and removed `int`. Could you review it?
I like your idea and I agree we can discuss the design as a group.

include/musica/micm.hpp
boulderdaze: Adding [link](https://github.com/NCAR/musica/actions/runs/9864306616/job/27238959697?pr=172) for the error messages

cmake/test_util.cmake
K20shores: I would like to keep fortran separated, but with the inclusion of tuvx into the musica-c library that line is blurred. Because of that I think we should go with Matt's suggestion. The other option is to check if TUVX is enabled before we check for OpenMP. If it is, enable the fortran language so that the fortran part of OpenMP is found, if that's possible

fortran/micm.F90
```diff
@@ -25,13 +26,22 @@ module musica_micm
       real(c_double)     :: final_time_ = 0._c_double
    end type solver_stats_t_c
 
+   ! We could use Fortran 2023 enum type feature if Fortran 2023 is supported
+   ! https://fortran-lang.discourse.group/t/enumerator-type-in-bind-c-derived-type-best-practice/5947/2
+   enum, bind(c)
+      enumerator :: Rosenbrock              = 1
+      enumerator :: RosenbrockStandardOrder = 2
+   end enum
```
K20shores, line 34: As long as our CI tests pass then I'm fine with this. The only potential issue maybe is the NAG compiler and if it supports this and if we need to compile with NAG ever.

But I like this as it is

include/musica/micm.hpp
```diff
@@ -162,20 +207,66 @@ namespace musica
...
+    std::unique_ptr<Rosenbrock> rosenbrock_;
 
-    std::unique_ptr<Rosenbrock> solver_;
+    /// @brief Standard-ordered Rosenbrock solver type
+    using DenseMatrixStandard = micm::Matrix<double>;
+    using SparseMatrixStandard = micm::SparseMatrix<double, micm::SparseMatrixStandardOrdering>;
+    using RosenbrockStandardType = typename micm::RosenbrockSolverParameters::
+        template SolverType<micm::ProcessSet, micm::LinearSolver<SparseMatrixStandard, micm::LuDecomposition>>;
+    using RosenbrockStandard = micm::Solver<RosenbrockStandardType, micm::State<DenseMatrixStandard, SparseMatrixStandard>>;
+    std::unique_ptr<RosenbrockStandard> rosenbrock_standard_;
```
K20shores, line 227: Are we allowed to have a pointer to a `micm::Solver` without specifying the template types?

src/micm/micm.cpp
```diff
@@ -167,21 +232,55 @@ namespace musica
...
-                                                 .Build());
+      rosenbrock_ = std::make_unique<Rosenbrock>(
+          micm::SolverBuilder<
+              micm::RosenbrockSolverParameters,
+              micm::VectorMatrix<double, MICM_VECTOR_MATRIX_SIZE>,
+              micm::SparseMatrix<double, micm::SparseMatrixVectorOrdering<MICM_VECTOR_MATRIX_SIZE>>,
+              micm::ProcessSet,
+              micm::LinearSolver<
+                  micm::SparseMatrix<double, micm::SparseMatrixVectorOrdering<MICM_VECTOR_MATRIX_SIZE>>,
+                  micm::LuDecomposition>>(micm::RosenbrockSolverParameters::ThreeStageRosenbrockParameters())
```
K20shores, line 251: maybe we should make something similar to `micm::CpuSolverBuilder` but make it be `micm::VectorizedCpuSolverBuilder` which is templated on the on vector size. Then it would be able to do all of this gross stuff for us. As an added benefit we should be able to add some alias types inside all the solver builder aliases that give us the type of the solver so that inside the header file we would only need to do something like

```
using builderType.= micm::CpuSolverBuilder;
using solverType = builderType::RosenbrockType;
```

K20shores (approved):
I think it all looks fine as is, but we could make it slightly better. I wonder if we could remove the need to have two member variable pointers to different solver types and two different build functions when only the matrix types are different

include/musica/micm.hpp
K20shores: or maybe we use `std::variant`?

fortran/micm.F90
boulderdaze: Jesse mentioned that it's probably not yet to use Fortran 2023 features but I'll leave the comment so we can revisit later.

include/musica/micm.hpp
boulderdaze: Do you mean something like `std::unique_ptr<micm::Solver> solver`? If it is the case, I tried it but it didn't work, (`template argument is invalid`). Unless`MICM` is a template class which is not a viable option, I think all the types must be specified.

I tried implementing `std::variant` and realized that you still need to do type check to get the data that variant holds, and the syntax can be quite tricky. For example, `MICM::Solve( lots of params ... )`
```c++
// Current
micm::State state = std::get<Rosenbrock>(solver)->GetState();

// std::variant
if(std::get_if<Rosenbrock>(&solver))
  micm::State state = std::get<Rosenbrock>(&solver)->GetState();
if(std::get_if<RosenbrockStandardOrder>(&solver))
  micm::State state = std::get<RosenbrockStandardOrder>(&solver)->GetState();
```

What do you think? The fact that C doesn't support any form of overloading, we quite have limitations here.

src/micm/micm.cpp
boulderdaze: I like this idea. I created [an issue](https://github.com/NCAR/micm/issues/602) for this in MICM. I think this can be a separate issue because alias-declaration for builder type should be done first at MICM.

cmake/test_util.cmake
boulderdaze: I created [a separate issue](https://github.com/NCAR/musica/issues/179) to tackle this.

fortran/micm.F90
mattldawson: I like this too

mattldawson (approved):
looks great!

include/musica/micm.hpp
mattldawson: that's cool, I think it's fine as-is also

include/musica/micm.hpp
K20shores: Hmm. Good point. What you have seems fine anyways

K20shores (approved):

PR #173, opened 2024-07-08
MUSICA Tutorial Chapter 0

dwfncar:
Added Chapter 0 tutorial describing a CMakeLists.txt to link to the musica-fortran library.


K20shores (approved):

tutorial/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+
+project(
+  musica-demo
+  VERSION 0.1
+  LANGUAGES CXX C Fortran
+)
+
+# mkdir build
+# cd build
+# cmake -DMUSICA_INSTALL_DIR=<musica_install_dir> ..
```
boulderdaze, line 11: Doesn't look like there is `MUSICA_INSTALL_DIR` option. Is it still valid? I also think it could be omitted since `chapter0.rst` covers the build process. What do you think?

tutorial/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+
+add_executable(demo_c demo.c)
+add_executable(demo_f demo.f90)
+
+target_include_directories(demo_c PUBLIC ${MUSICA_INCLUDE_DIR})
+target_link_directories(demo_c PUBLIC ${MUSICA_LIB_DIR})
+
+target_include_directories(demo_f PUBLIC ${MUSICA_INCLUDE_DIR})
+target_link_directories(demo_f PUBLIC ${MUSICA_LIB_DIR})
+target_link_libraries(demo_f musica-fortran musica stdc++)
```
boulderdaze, line 27: Instead of using `target_link_directories`, could you use `target_link_libraries` as it is discouraged to use as stated in the docs? I also wonder whether you need them. When you just use `target_link_librararies` to link musica, does it fail?

boulderdaze (changes requested):
Looks good! I just left a couple questions.

tutorial/CMakeLists.txt
mattldawson: yes, I would add any tutorial code to the test suite

boulderdaze: Great, @dwfncar, could you make `demo_*` as a test?

tutorial/CMakeLists.txt
dwfncar: MUSICA_INSTALL_DIR is not defined during the MUSICA configuration, but rather during the host model configuration via the cmake -D option, and points to the installation location of the MUSICA library.

tutorial/CMakeLists.txt
dwfncar: Without the `target_link_libraries` I get the link errors `/usr/bin/ld: cannot fine -lmusica-fortran` (and -`lmusica`)

dwfncar: Added demo.f90 as a unit test in fortran/test/unit/CMakeLists.txt, will add demo.c when it has actual content.

tutorial/CMakeLists.txt
boulderdaze: Ah, I see. I didn't know that

tutorial/CMakeLists.txt
boulderdaze: hmm maybe instead of`tutorial/` being the top level directory it might need to be a subdirectory of`fortran/` like `micm`? Then it could link to `musica-fortran` library. What do you think?

mattldawson: Yeah, I would recommend the tutorial tests being in `micm/fortran/test/tutorial` and you should be able to just use the `create_standard_fortran_test()` macro (I think). That macro links tests against the `musica-fortran` library. Similar to this: https://github.com/NCAR/musica/blob/69bcb2efed80afab88be74809a44df43717b9a73/fortran/test/unit/CMakeLists.txt#L6-L8

boulderdaze (approved):
Looks good to me! Thank you for addressing the comments!

fortran/test/tutorial/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+set(MUSICA_LIB_DIR "${MUSICA_INSTALL_DIR}/lib")
+
+message(STATUS "${MUSICA_INCLUDE_DIR}")
+message(STATUS "${MUSICA_LIB_DIR}")
+
+add_executable(demo_f demo.f90)
+
+target_include_directories(demo_f PUBLIC ${MUSICA_INCLUDE_DIR})
+target_link_directories(demo_f PUBLIC ${MUSICA_LIB_DIR})
+target_link_libraries(demo_f musica-fortran musica stdc++)
```
boulderdaze, line 23: Oops sorry, I missed it, could you remove `target_link_directories(..)`?
I also think you don't need to link to the musica library.
```suggestion
target_include_directories(demo_f PUBLIC ${MUSICA_INCLUDE_DIR})
target_link_libraries(demo_f PRIVATE musica-fortran)
```

boulderdaze (commented):
Sorry, I missed `target link directory` thing before I submitted the reveiw

PR #181, opened 2024-07-23
Python interface for TUVX functions. 

WardF:
Draft PR so that I can keep an eye on any issues that creep in.


WardF:
Re-implementing some code that didn't get migrated the way I thought it did from my initial sandbox.

zmoon:
This doesn't seem to have gotten into main. Is there another effort?

K20shores:
@zmoon not as of now. We already have a bunch of C APIs that we could use to call into TUVx. We have to wrap those with pybind11 and then figure out how to distribute them across all operating systems. Windows makes that hard because we need a Fortran compiler.

Eventually, TUVx will make it into Python, but it's not a high priority for us right now. We also have another effort starting up that will optimize TUVx because it can be slow, and make it available for solving on GPUs. We will be refactoring TUVx into C++ so that a Fortran compiler, except for regression tests, will no longer be needed.

PR #184, opened 2024-08-05
159 create an action to publish to pypi when we make a release

### LINKED ISSUE #159: Create an action to publish to pypi when we make a release
Add an action that publishes the MUSICA Python wrapper to pypi for each release.

**Acceptance Criteria**
- The MUSICA Python package is deployed to pypi on each commit to the release branch

**Ideas**
- Follow the logic [here](https://github.com/pybind/cmake_example/blob/master/.github/workflows/wheels.yml) and [here](https://github.com/pybind/cmake_example/blob/7a94877f581a14de4de1a096fb053a55fc2a66bf/pyproject.toml#L33-L38)
- Apparently, can set up [publishers](https://docs.pypi.org/trusted-publishers/) with pypi so that we don't need to store a token
- Investigate whether it's possible to have the version number in just one place, instead of in both the CMakeLists.txt file and pyproject.toml

K20shores:
Closes #159 

- uses [cbuildwheel](https://github.com/pypa/cibuildwheel) to cross-compile our python library for many architectures
- setup our pypi repository as a [trusted publisher](https://pypi.org/manage/project/musica/settings/publishing/#errors) so that we don't need to store a github secret

Proof that it works, but these will be deleted. [Successful CI run that built the most recent of these](https://github.com/NCAR/musica/actions/runs/10253565114)

Screenshot 2024-08-05 at 1 32 39 PM

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -47,7 +47,7 @@ jobs:
 
     - uses: pypa/cibuildwheel@v2.17
       env:
-        CIBW_ARCHS_MACOS: arm64 x86_64
+        CIBW_ARCHS_MACOS: arm64 
         CIBW_SKIP: cp27-* cp34-* cp35-* cp36-*
         CIBW_BUILD: cp37-* cp38-* cp39-* cp310-* cp311-* cp312-*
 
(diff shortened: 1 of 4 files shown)


WardF (approved):

boulderdaze (approved):
Nice!

mattldawson (approved):
nice!

Issue #178, opened 2024-07-11
Use alias-declaration for vector-ordered Rosenbrock solver 


boulderdaze:
This task can bet started after [Create alias-declaration for vector-ordered solver builder](https://github.com/NCAR/micm/issues/602) is completed in MICM

boulderdaze:
Resolved

Issue #156, opened 2024-06-21
Investigate convergence failure by Rosenbrock solver using Chapman 


boulderdaze:
`musica::SolverStats` using Chapman returns the following profile indicating convergence failure

```c++
Solver state: Convergence Exceeded Max Steps
Function Calls: 1997
Jacobian updates: 996
Number of steps: 1001
Accepted: 996
Rejected: 1
Decompositions: 1001
Solves: 3003
Singular: 0
Final time: 0.290802
```
After M is labeled as "THIRD_BODY"
```c++
Solver state: Convergence Exceeded Max Steps
Function Calls: 1995
Jacobian updates: 994
Number of steps: 1001
Accepted: 994
Rejected: 1
Decompositions: 1001
Solves: 3003
Singular: 0
Final time: 0.246266
```
**Acceptance Criteria**
Should achieve a converged solution
- MUSICA example configuration should use `THIRD_BODY` species for `M`

**Ideas**
The tests might have to be updated to set the air density
-  "M" species needs to be set to the new "THIRD_BODY" type

boulderdaze:
Specifying `THIRD_BODY` for `M` doesn't fix the issue.

PR #192, opened 2024-08-08
Add multi-grid cell solving to MUSICA API for MICM

### LINKED ISSUE #175: Allow solving of multiple grid cells in MICM
Users should be able to solve multiple grid cells at the same time using MICM.

**Acceptance Criteria**
- Users can use any MICM solver (Standard/Vector Rosenbrock, Backward Euler, etc.) to solve multiple grid cells
- Users can update initial species concentrations for all grid cells prior to calling Solve
- Users can update environmental conditions (temperature, pressure, air density) for all grid cells prior to calling Solve
- Users can update custom rate constant parameters for all grid cells prior to calling Solve
- If using a vector matrix solver, users should be able to pass in data in the "standard" matrix format and the MUSICA function should remap it to the vector matrix ordering, including when the number of grid cells is not evenly divisible by the vector dimension

mattldawson:
Adds the ability to solve multiple grid cells with a single call to the MICM solver. Includes updates to the Fortran and Python wrappers. Also, modifies the Chapman test conditions to avoid convergence failures.

closes #175 

I don't really like the fact that these function calls accept pointers to 2D data for temperature, pressure, concentrations, etc. where the size and shape are assumed to be correct with no checks. I suggest that we create a separate issue to address this.


fortran/micm.F90
```diff
@@ -51,23 +51,21 @@ subroutine delete_micm_c(micm, error) bind(C, name="DeleteMicm")
...
-      type(solver_stats_t_c), intent(out)    :: solver_stats
-      type(error_t_c), intent(inout)         :: error
+      type(c_ptr),         value, intent(in)    :: micm
+      real(kind=c_double), value, intent(in)    :: time_step
+      type(c_ptr),         value, intent(in)    :: temperature
+      type(c_ptr),         value, intent(in)    :: pressure
+      type(c_ptr),         value, intent(in)    :: air_density
+      type(c_ptr),         value, intent(in)    :: concentrations
+      type(c_ptr),         value, intent(in)    :: user_defined_reaction_rates
+      type(string_t_c),           intent(out)   :: solver_state
```
boulderdaze, line 66: Doesn't `concentrations` have to be `intent(inout)`?
```fortran
type(c_ptr), value, intent(inout) :: concentrations
```

configs/chapman/reactions.json
```diff
@@ -14,7 +14,7 @@
                             "yield": 2.0
                         }
                     },
-                    "MUSICA name": "R1"
+                    "MUSICA name": "jO2"
```
boulderdaze, line 17: Why are those names changed?

boulderdaze (approved):
I really like what you did! Your PR closes [#156](https://github.com/NCAR/musica/issues/156) and partially completes [#178](https://github.com/NCAR/musica/issues/178) (maybe this is enough?)
Your test coverage looks impressive. Looks like incorrect mapping between the species and concentrations caused the convergence failure, did it?

fortran/micm.F90
boulderdaze: oh it's just `type(c_ptr)`, so probably not

K20shores (approved):

configs/analytical/reactions.json
```diff
@@ -7,14 +7,26 @@
...
+          "type": "USER_DEFINED",
+          "MUSICA name" : "reaction 2",
+          "reactants": {"E": {"qty": 1}},
+          "products": {"F": {"yield": 1}}
+        },
+        {
+          "type": "USER_DEFINED",
+          "MUSICA name" : "reaction 1",
+          "reactants": {"D": {"qty": 1}},
+          "products": {"E": {"yield": 1}}
```
K20shores, line 19: My only concern with this is that `USER_DEFINED` isn't technically part of the camp configuration. However, I do think this is incredibly useful. We could remove photolysis, first order loss, emission, from the technical specification of the configuration because they are all just logical versions of user defined and it allows us to not have to do funny things with music box for integrated reaction rates.

I guess there's nothing actionable with that paragraph; I'm just putting thoughts out there

fortran/micm.F90
mattldawson: Yeah, I think this is just saying the address pointed to doesn't change

mattldawson:
> I really like what you did! Your PR closes [#156](https://github.com/NCAR/musica/issues/156) and partially completes [#178](https://github.com/NCAR/musica/issues/178) (maybe this is enough?) Your test coverage looks impressive. Looks like incorrect mapping between the species and concentrations caused the convergence failure, did it?

Thanks! I added #156 to the issues this closes. I think it might be enough for #178 also. If you agree, I can include it in the list.

boulderdaze:
> > I really like what you did! Your PR closes [#156](https://github.com/NCAR/musica/issues/156) and partially completes [#178](https://github.com/NCAR/musica/issues/178) (maybe this is enough?) Your test coverage looks impressive. Looks like incorrect mapping between the species and concentrations caused the convergence failure, did it?
> 
> Thanks! I added #156 to the issues this closes. I think it might be enough for #178 also. If you agree, I can include it in the list.

Sounds good, I agree with you

dwfncar (approved):
Will update 162_euler_option branch with the changes to MicmSolve.

PR #214, opened 2024-08-30
Allow querying of grid size

mattldawson:
Adds the ability to return the number of sections in a grid

### DIFF
--- a/fortran/tuvx/grid.F90
+++ b/fortran/tuvx/grid.F90
@@ -154,7 +154,7 @@ contains
   !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
   
   integer function number_of_sections(this, error) result( n_sections )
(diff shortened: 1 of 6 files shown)


fortran/tuvx/grid.F90
```diff
@@ -140,6 +151,23 @@ function grid_t_constructor(grid_name, grid_units, number_of_sections, error) &
 
   end function grid_t_constructor
   
+  !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+  
+  integer function number_of_sections(this, error) result( n_sections )
+    use musica_util, only: error_t, error_t_c, dk => musica_dk
+
```
boulderdaze, line 158: Looks like `dk` is not used. Could you remove this?

boulderdaze (changes requested):
Looks good! Just a very minor request

fortran/tuvx/grid.F90
mattldawson: done!

boulderdaze (approved):

PR #223, opened 2024-09-30
Add TUV-x run function

mattldawson:
Adds the ability to run the TUV-x calculations through the C and Fortran APIs. This leaves the TUV-x Python wrapper for future work.


K20shores (approved):

include/musica/tuvx/tuvx.hpp
```diff
@@ -66,15 +76,19 @@ namespace musica
     GridMap *GetGridMap(TUVX *tuvx, Error *error);
     ProfileMap *GetProfileMap(TUVX *tuvx, Error *error);
     RadiatorMap *GetRadiatorMap(TUVX *tuvx, Error *error);
+    void RunTuvx(TUVX *tuvx, const double solar_zenith_angle, const double earth_sun_distance,
+                 double * const photolysis_rate_constants, double * const heating_rates, Error * const error);
```
boulderdaze, line 80: I guess `const pointer to Error` is fine, but only this `Error` parameter is marked by `const` compared to the other functions that take `Error`
```C++
Error * const error
```

src/tuvx/tuvx.cpp
```diff
@@ -152,4 +159,17 @@ namespace musica
     return radiator_map;
   }
 
+  void TUVX::Run(const double solar_zenith_angle, const double earth_sun_distance,
+                 double * const photolysis_rate_constants, double * const heating_rates, Error * const error)
+  {
+    *error = NoError();
+    int error_code = 0;
+    double sza_degrees = solar_zenith_angle * 180.0 / M_PI;
```
boulderdaze, line 167: Can't find the definition of `M_PI` within the repository. Where is it defined?

valgrind.supp
```diff
@@ -5,19 +5,11 @@
...
-   Memcheck:Leak
-   match-leak-kinds: possible
-   fun:malloc
+   <DiagnosticConditionalJump>
+   Memcheck:Cond
+   fun:__tuvx_diagnostic_util_MOD_diagnostic_1d_dk
+   fun:__tuvx_photolysis_rates_MOD_get
+   fun:__tuvx_core_MOD_run
+   fun:InternalRunTuvx
    ...
```
boulderdaze, line 14: Is this necessary?

boulderdaze (changes requested):
Looks good! What are the differences between the fixed and the original configuration?

include/musica/tuvx/tuvx.hpp
```diff
@@ -48,10 +48,20 @@ namespace musica
     /// @return a radiator map pointer
     RadiatorMap *CreateRadiatorMap(Error *error);
 
+    /// @brief Run the TUV-x photolysis calculator
+    /// @param solar_zenith_angle Solar zenith angle [radians]
+    /// @param earth_sun_distance Earth-Sun distance [AU]
+    /// @param photolysis_rate_constants Photolysis rate constant layer and reaction [s^-1]
```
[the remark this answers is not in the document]
mattldawson: fixed!

include/musica/tuvx/tuvx.hpp
mattldawson: Maybe we could make a follow-up task to go through all the code and use `const` consistently and where it makes sense?

src/tuvx/tuvx.cpp
mattldawson: Well, I thought it was part of the C standard but apparently it's not, it's just commonly included in `math.h`. I switched this to use `std::numbers::pi` instead

include/musica/tuvx/tuvx.hpp
boulderdaze: Yeah, I like the idea

valgrind.supp
mattldawson: Valgrind was catching a conditional jump on uninitialized data that I couldn't really make sense of because the source code line that seemed to trigger it was a fortran function header. it seems to only be related to when we try to output diagnostics inside TUV-x, so I figured it wouldn't be a big deal to suppress the error. But, I could also make a separate issue to look into this.

boulderdaze: Ah I see. Yeah I agree. I don't think it wouldn't be a big deal for that case

mattldawson:
> Looks good! What are the differences between the fixed and the original configuration?

The fixed configuration is where everything is specified in the json file, and the other test includes setting some grids and profiles through the API. Is this what you were asking?

boulderdaze:
> > Looks good! What are the differences between the fixed and the original configuration?
> 
> The fixed configuration is where everything is specified in the json file, and the other test includes setting some grids and profiles through the API. Is this what you were asking?

oh I was curious what's fixed in the new fixed configuration that are different from the old(?) configuration

mattldawson:
> > > Looks good! What are the differences between the fixed and the original configuration?
> > 
> > 
> > The fixed configuration is where everything is specified in the json file, and the other test includes setting some grids and profiles through the API. Is this what you were asking?
> 
> oh I was curious what's fixed in the new fixed configuration that are different from the old(?) configuration

ah, I think it's pretty much the same as the old configuration. I just renamed it fixed because I added this other configuration where some of the data is changed by the user

Issue #217, opened 2024-09-10
Provide an interface to the micm State


K20shores:
In #225, we started storing a single state. This is because large systems take a long time (~5 seconds) to initialize the state. Repeated calls to solve from python were constantly making and deleting a state which very drastically degraded solving time. 

Additionally, fortran and python have to call solve with many arguments

https://github.com/NCAR/musica/blob/954265b2311ce4037086ddbc4f2ac0739ec1190c/fortran/test/fetch_content_integration/test_micm_api.F90#L122-L123

https://github.com/NCAR/musica/blob/954265b2311ce4037086ddbc4f2ac0739ec1190c/python/test/test_analytical.py#L55-L62

And, when that data gets into C++, we have to copy it onto the state

https://github.com/NCAR/musica/blob/326b5119768d5be9654baf96ae3bd6a1b757fdc8/src/micm/micm.cpp#L437-L450

https://github.com/NCAR/musica/blob/326b5119768d5be9654baf96ae3bd6a1b757fdc8/python/wrapper.cpp#L48-L122

This means that we are doing many copies of the data before we even get to solve.

Write a wrapper around the micm state class which allows for data to be accessed and set directly in both python and fortran.

**Acceptance criteria**
- A new State wrapper class is created and accessible everywhere
- A state is not stored alongside a solver
  - https://github.com/NCAR/musica/blob/326b5119768d5be9654baf96ae3bd6a1b757fdc8/include/musica/micm.hpp#L254-L267
- All solving interfaces are updated to pass a state
  - You may be able to make a solver state pair and just pass that as we do now, but be able to remove the other arguments. Note that time_step is not a state parameter
  - https://github.com/NCAR/musica/blob/326b5119768d5be9654baf96ae3bd6a1b757fdc8/src/micm/micm.cpp#L418-L428

**Ideas**
- We need to figure out how [this issue in micm](https://github.com/NCAR/micm/issues/625) gets resolved for us to really figure this out
  - That did get resolved and it was a bug in micm.
- Add a new State class in https://github.com/NCAR/musica/blob/main/include/musica/micm.hpp or in a new file
  - support operations to
    - get and set temperature
    - get and set pressure
    - get and set air density
    - get and set concentrations
    - set user defined reaction rates
- The implementation of all of the above can go in https://github.com/NCAR/musica/blob/main/src/micm/micm.cpp, or you can make a new file
[...]

PR #225, opened 2024-10-02
Store a state alongside the solver

### LINKED ISSUE #224: Python solving is slow
Guy and @sancyw have found that using music box interactive is slow after we moved from CAMP to MICM.

After investigating, I've found that creating the state is prohibitively slow for TS1 when the integrated reaction rate species are added (about 5 seconds to create the state).

Right now, we make a state each time we call solve. But, with new updates from micm, we can save and reuse the same state. 

**Acceptance criteria**
- One state is created for a solver and reused

**Ideas**
- Store a `std::pair` of solver type and state

### LINKED ISSUE #244: CAM-SIMA Documentation
Do we have enough documentation to show how to build, configure, and use MUSICA?

We would need to ensure that there is enough documentation to:

- create a mechanism for micm
- create a mechanism for tuvx
- create a link for tuvx->micm
- use the configuration

I think that our documentation for musica, tuvx, and musicbox aren't all updating properly and that only micm is. I have not verified this

K20shores:
Closes #224

Using this version of musica in music box makes the slow example Sancy pointed out extremely fast to run, though the construction still takes about 5 seoncds

This is a temporary fix. #217 will provide an interface to the state that python could hold and use directly

### DIFF
--- a/include/musica/micm.hpp
+++ b/include/musica/micm.hpp
@@ -17,6 +17,8 @@
 #include <micm/util/sparse_matrix_vector_ordering.hpp>
 #include <micm/util/vector_matrix.hpp>
 
+#include <utility>
+#include <chrono>
 #include <cstddef>
 #include <map>
 #include <memory>
@@ -61,8 +63,6 @@ namespace musica
       int64_t decompositions_{};
       /// @brief The number of linear solves
       int64_t solves_{};
-      /// @brief The number of times a singular matrix is detected.
-      int64_t singular_{};
       /// @brief The final time the solver iterated to
       double final_time_{};
       /// @brief The final state the solver was in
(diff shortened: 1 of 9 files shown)


mattldawson (approved):
awesome!

include/musica/micm.hpp
```diff
@@ -201,7 +198,7 @@ namespace musica
     /// @param custom_rate_parameters Array of custom rate parameters [grid cell][parameter] (various units)
     /// @param error Error struct to indicate success or failure
     void Solve(
-        auto &solver,
+        auto &solver_state_pair,
```
boulderdaze, line 201: Could you update the `line 192` so it can include the description for the state?

```C++
    /// @param solver Pointer to solver
```

boulderdaze (approved):
Looks good!

PR #226, opened 2024-10-02
Add Cloud-J to MUSICA

lizziel:
### Name and Institution (Required)

Name: Lizzie Lundgren
Institution: Harvard University

### Describe the update

This draft PR adds the [Cloud-J photolysis package](https://github.com/geoschem/Cloud-J) to MUSICA. This is in the very early stages and it does not yet successfully build. Comments/discussion welcome!

### Expected changes

This should be a zero diff update for MUSICA if Cloud-J build is disabled, which is what I set as the default. Note, however, that I currently changed the defaults to build the fortran interface. Previously that was off by default.

### Reference(s)

This version uses Cloud-J 8.0.1: https://zenodo.org/records/13862693

### Related Github Issue

None


lizziel:
Note that the fortran subdirectory is currently populated by tuvx files that I renamed and edited to be Cloud-J instead. Obviously this is wrong but it is a first pass attempt to compile.

lizziel:
I am closing this PR since we are planning to link Cloud-J in MUSICA Fortran Library only. I will make future PRs from branches on github.com/ncar/music. @mattldawson, could you add me as collaborator?

PR #233, opened 2024-10-09
fix bug in getting number of grid sections

mattldawson:
Fixes a bug with the pointer used to access the number of grid sections and adds some missing test coverage


K20shores (approved):

src/tuvx/interface_grid.F90
```diff
@@ -117,23 +117,23 @@ end subroutine internal_delete_grid_updater
 
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
-  function internal_get_num_sections(grid, error_code) &
+  function internal_get_num_sections(updater, error_code) &
       bind(C, name="InternalGetNumSections") result(num_sections)
     use iso_c_binding, only: c_ptr, c_f_pointer, c_int, c_size_t
-    use tuvx_grid_from_host, only: grid_from_host_t
+    use tuvx_grid_from_host, only: grid_from_host_t, grid_updater_t
```
boulderdaze, line 124: Can we remove this?
```suggestion
    use tuvx_grid_from_host, only: grid_updater_t
```

boulderdaze (approved):
Looks good! Just a single minor request

src/tuvx/interface_grid.F90
mattldawson: yes, removed!

PR #238, opened 2024-11-04
Use fortran types in API functions

mattldawson:
Updates fortran API function in MUSICA library to accept fortran floating point and integer types. Also overloads the MICM `solve()` function in the fortran API to allow users to pass C pointers to the input/output arrays. This is intended to reduce the number of copies, but requires the user to be responsible for ordering the matrix data correctly for the vector/standard matrix types.

These updates are temporary until #217 is done.


fortran/test/fetch_content_integration/test_micm_api.F90
```diff
@@ -36,38 +37,38 @@ program test_micm_api
...
+    type(micm_t), pointer         :: micm
+    real(real64)                  :: time_step
+    real(real64)                  :: temperature(1)
+    real(real64)                  :: pressure(1)
+    real(real64)                  :: air_density(1)
+    real(real64), dimension(4)    :: concentrations 
+    real(real64), dimension(3)    :: user_defined_reaction_rates 
+    character(len=256)            :: config_path
+    integer(c_int)                :: solver_type
+    integer(c_int)                :: num_grid_cells
```
boulderdaze, line 60: Would it be just `integer`?

fortran/test/fetch_content_integration/test_micm_api.F90
```diff
@@ -36,38 +37,38 @@ program test_micm_api
...
+    character(len=:), allocatable :: string_value
+    real(real64)                  :: double_value
+    integer(c_int)                :: int_value
```
boulderdaze, line 63: Would it be just integer?

fortran/test/fetch_content_integration/test_micm_api.F90
```diff
@@ -172,27 +173,27 @@ subroutine test_multiple_grid_cells(micm, NUM_GRID_CELLS, time_step, test_accura
...
-    real(c_double), target    :: user_defined_reaction_rates(NUM_GRID_CELLS * NUM_USER_DEFINED_REACTION_RATES)
+    real(real64), target    :: temperature(NUM_GRID_CELLS)
+    real(real64), target    :: pressure(NUM_GRID_CELLS)
+    real(real64), target    :: air_density(NUM_GRID_CELLS)
+    real(real64), target    :: concentrations(NUM_GRID_CELLS * NUM_SPECIES)
+    real(real64), target    :: initial_concentrations(NUM_GRID_CELLS * NUM_SPECIES)
+    real(real64), target    :: user_defined_reaction_rates(NUM_GRID_CELLS * NUM_USER_DEFINED_REACTION_RATES)
     type(string_t)            :: solver_state
     type(solver_stats_t)      :: solver_stats
     integer(c_int)            :: solver_type
```
boulderdaze, line 189: Would it be just `integer`? 
```Fortran
integer :: solver_type
```
It's a minor request, but could we use consistent spacing for the listed variables to improve visual clarity?

fortran/test/fetch_content_integration/test_micm_box_model.F90
```diff
@@ -19,14 +20,14 @@ subroutine box_model()
...
-    real(c_double), target :: pressure(1)
-    real(c_double), target :: air_density(1)
-    real(c_double), target :: concentrations(6)
-    real(c_double), target :: user_defined_reaction_rates(2)
+    real(real64) :: time_step
+    real(real64), target :: temperature(1)
+    real(real64), target :: pressure(1)
+    real(real64), target :: air_density(1)
+    real(real64), target :: concentrations(6)
+    real(real64), target :: user_defined_reaction_rates(2)
```
boulderdaze, line 30: Should we remove target attributes?

boulderdaze (changes requested):

fortran/test/fetch_content_integration/test_micm_api.F90
mattldawson: fixed!

fortran/test/fetch_content_integration/test_micm_api.F90
mattldawson: fixed!

fortran/test/fetch_content_integration/test_micm_api.F90
mattldawson: fixed, and reformatted

fortran/test/fetch_content_integration/test_micm_box_model.F90
mattldawson: I think these have to be `target` now, because the `solve` function expects arrays with the `target` attribute because it uses `c_loc` on them

boulderdaze: Ah I missed that. Sorry!

K20shores (approved):

fortran/micm.F90
```diff
@@ -135,10 +135,14 @@ end function get_user_defined_reaction_rates_ordering_c
   type :: micm_t
     type(mappings_t), pointer :: species_ordering => null()
     type(mappings_t), pointer :: user_defined_reaction_rates => null()
-    type(c_ptr), private         :: ptr = c_null_ptr
+    type(c_ptr), private      :: ptr = c_null_ptr
+    integer,     private      :: number_of_grid_cells = 0
+    integer,     private      :: solver_type = 0
```
boulderdaze, line 140: Would it make more sense to set them to a default value rather than an invalid value? My main concern is that `solver_type` is an enum, and we expect the input to be a string that maps to the corresponding integer value. Setting them as integers could give an impression that the type is actually an integer. 
```Fortran
  enum, bind(c)
    enumerator :: Rosenbrock                 = 1
    enumerator :: RosenbrockStandardOrder    = 2
    enumerator :: BackwardEuler              = 3
    enumerator :: BackwardEulerStandardOrder = 4
  end enum
```

fortran/test/fetch_content_integration/test_micm_box_model.F90
```diff
@@ -3,30 +3,33 @@ program test_micm_box_model
...
 
 contains
 
-  subroutine box_model()
+  !> Runs a simple box model using the MICM solver and passing in fortran arrays
+  subroutine box_model_arrays()
 
     character(len=256) :: config_path
     integer(c_int)     :: solver_type
     integer(c_int)     :: num_grid_cells
```
boulderdaze, line 23: Would they be just `integer`?

fortran/test/fetch_content_integration/test_micm_box_model.F90
```diff
@@ -63,6 +66,61 @@ subroutine box_model()
...
 
-  end subroutine box_model
+  end subroutine box_model_arrays
+
+  !> Runs a simple box model using the MICM solver and passing in C pointers
+  subroutine box_model_c_ptrs()
+
+    character(len=256) :: config_path
+    integer(c_int)     :: solver_type
+    integer(c_int)     :: num_grid_cells
```
boulderdaze, line 76: Would they be just `integer`?

boulderdaze (approved):
Looks good! Just a couple minor questions, but they shouldn't prevent your PR from being merged.

fortran/micm.F90
mattldawson: good point! I added an `UndefinedSolver` option for the initial value

fortran/test/fetch_content_integration/test_micm_box_model.F90
mattldawson: fixed!

fortran/test/fetch_content_integration/test_micm_box_model.F90
mattldawson: fixed!

PR #242, opened 2024-11-06
Add `contiguous` attribute for arrays passed to c functions

mattldawson:
Adds the `contiguous` attribute to fortran API function array arguments that are passed as C-style pointers to MUSICA library C functions. This allows strided-access array slices to be passed to these functions, but still pass contiguous-in-memory arrays to the C functions. Updates related tests to pass array slices from fortran.


K20shores (approved):

boulderdaze (approved):
Looks good!

PR #250, opened 2025-02-13
Add size function to index mappings

mattldawson:
Adds a function that returns the number of elements in an index mappings object.

### DIFF
--- a/fortran/util.F90
+++ b/fortran/util.F90
@@ -147,6 +147,7 @@ module musica_util
   private
     type(index_mappings_t_c) :: mappings_c_
   contains
+    procedure :: size => index_mappings_size
     procedure :: copy_data
     final :: index_mappings_finalize
   end type index_mappings_t
@@ -219,6 +220,12 @@ module musica_util
       import :: index_mappings_t_c
       type(index_mappings_t_c), intent(inout) :: mappings
     end subroutine delete_index_mappings_c
+    function get_index_mappings_size_c( mappings ) result( size ) &
+        bind(c, name="GetIndexMappingsSize")
+      import :: index_mappings_t_c, c_size_t
+      type(index_mappings_t_c), value, intent(in) :: mappings
+      integer(c_size_t) :: size
+    end function get_index_mappings_size_c
     subroutine copy_data_c(mappings, source, target) bind(c, name="CopyData")
       import :: index_mappings_t_c, c_ptr
       type(index_mappings_t_c), value, intent(in) :: mappings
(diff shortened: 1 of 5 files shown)


include/musica/util.hpp
```diff
@@ -131,6 +131,11 @@ namespace musica
     IndexMappings
     CreateIndexMappings(const Configuration configuration, const Mappings source, const Mappings target, Error* error);
 
+    /// @brief Returns the number of elements in an array of IndexMappings
+    /// @param mappings The array of IndexMappings
+    /// @return The number of elements
+    std::size_t GetIndexMappingsSize(const IndexMappings mappings);
+
```
boulderdaze, line 138: Could you explain the function parameter, the array of IndexMappings? The type is a struct, and I'm confused about what it means to have an array of IndexMappings. Since the function is returning the size, is it possible to declare the parameter as a const ref?

fortran/util.F90
```diff
@@ -219,6 +220,12 @@ pure subroutine delete_index_mappings_c( mappings ) &
       import :: index_mappings_t_c
       type(index_mappings_t_c), intent(inout) :: mappings
     end subroutine delete_index_mappings_c
+    function get_index_mappings_size_c( mappings ) result( size ) &
+        bind(c, name="GetIndexMappingsSize")
+      import :: index_mappings_t_c, c_size_t
+      type(index_mappings_t_c), value, intent(in) :: mappings
+      integer(c_size_t) :: size
+    end function get_index_mappings_size_c
     subroutine copy_data_c(mappings, source, target) bind(c, name="CopyData")
```
boulderdaze, line 229: Would it be possible to add a new line between the function definitions to improve readability?

boulderdaze (changes requested):
Looks good! I just have a couple minor questions.

K20shores (approved):

fortran/util.F90
mattldawson: done!

include/musica/util.hpp
mattldawson: Yeah, I agree the description was misleading. I revised the text, see if you think it is more accurate now. I left this as const instead of const ref, just to be consistent with all the other functions in this header, and because it's in an `extern "C"` block. Passing this by reference from Fortran would mean passing it as a pointer. But, the struct only has two elements, a pointer to the array of `IndexMapping` structs and the size. Let me know if you still think we should change it, and I can change the other functions also.

boulderdaze: Yeah that makes sense to me, thank you for addressing the questions.

boulderdaze (approved):

PR #251, opened 2025-02-18
Reorganize the MUISICA include files upon installation

boulderdaze:
-  Reconfigured cmake to include the Fortran module files only once and removed duplicates during installation
- Created micm directories to match the same structure as the `src/micm/micm.cpp` directory as well as `tuvx` setup
```CPP
 include/musica/micm.hpp  ->  include/musica/micm/micm.hpp
 fortran/micm.F90         ->  fortran/micm/micm.90 
```
Fortran module files are installed at `include/musica/fortran/`
```
include/
├── micm
│   ├── configure
│   ├── cuda
│   ├── ...
│   └── version.hpp
└── musica
    ├── component_versions.hpp
    ├── fortran
    │  ├── musica_array.mod
    │  ├── ...
    │  ├── tuvx_core.mod
    │  ├── ...
    ├── micm
    ├── tuvx
    ├── util.hpp
    └── version.hpp
```


K20shores (approved):

src/packaging/CMakeLists.txt
```diff
@@ -69,17 +45,15 @@ if (MUSICA_ENABLE_TUVX)
...
-      ${MUSICA_FORTRAN_SRC_DIR}/tuvx/profile_map.F90
-      ${MUSICA_FORTRAN_SRC_DIR}/tuvx/radiator.F90
-      ${MUSICA_FORTRAN_SRC_DIR}/tuvx/radiator_map.F90
-      ${MUSICA_FORTRAN_SRC_DIR}/tuvx/tuvx.F90
-    DESTINATION 
+    DIRECTORY
+      ${MUSICA_MOD_DIR}/../include/
+    DESTINATION
       ${MUSICA_INSTALL_INCLUDE_DIR}/musica/fortran
+    FILES_MATCHING PATTERN "*.mod"
```
boulderdaze, line 56: I realized that the tuvx code is written in fortran, and musica actually includes the tuvx object files as sources when building the musica library. So it makes sense that tuvx fortran mod files are built and installed along with musica. 
The musica-fortran build process only compiles and installs the source files located in its directory.

fortran/CMakeLists.txt
```diff
@@ -26,25 +27,20 @@ endif()
 add_library(musica-fortran)
 add_library(musica::musica-fortran ALIAS musica-fortran)
 
-if(NOT BUILD_SHAREDS_LIBS)
-  set_target_properties(musica-fortran
-    PROPERTIES
-    POSITION_INDEPENDENT_CODE ON
-  )
-endif()
-
```
boulderdaze, line 35: Hi @K20shores, I removed the position-independent property because I believe it's meant for shared libraries, and so far, `lmusica-fortran` is a static library. I wanted to check with you to see if you think this property should be kept.

K20shores: I don't remember why we needed this option. If we need it again we can always add it back; I'm fine with removing it for now

K20shores (approved):
I tested it out and it all seems to work for me

mattldawson (approved):
very nice!

PR #252, opened 2025-02-19
97 modify windows GitHub action to build and test the fortran code

### LINKED ISSUE #97: Modify windows github action to build and test the fortran code
Options
- mingw
- llvm clang

**Acceptance**
- the fortran options can be built on windows
- python can be installed on windows WITH TUVx options enabled

K20shores:
Closes #97

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -123,7 +123,7 @@ jobs:
       - name: Test
         run: cd build ; ctest -C Debug --output-on-failure --verbose
 
-  gnu:
+  fortran:
     runs-on: windows-latest
     strategy:
       fail-fast: false
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -160,7 +160,8 @@ jobs:
                               -D MUSICA_GIT_TAG=${{ github.sha }} \
                               -D MUSICA_ENABLE_MICM=ON \
                               -D MUSICA_ENABLE_TUVX=ON \
-                              -D MUSICA_ENABLE_PYTHON_LIBRARY=ON
+                              -D MUSICA_ENABLE_PYTHON_LIBRARY=ON \
+                              -D MUSICA_BUILD_FORTRAN_INTERFACE=ON
 
       - name: Build the project
         shell: msys2 {0}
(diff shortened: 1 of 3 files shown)


mattldawson (approved):
nice!

.github/workflows/windows.yml
```diff
@@ -122,3 +122,51 @@ jobs:
...
+      - name: Configure with CMake
+        shell: msys2 {0}
+        run: |
+          cmake -S . -B build -G "MinGW Makefiles" \
+                              -D CMAKE_BUILD_TYPE=${{ matrix.build_type }} \
+                              -D MUSICA_GIT_TAG=${{ github.sha }} \
+                              -D MUSICA_ENABLE_MICM=ON \
+                              -D MUSICA_ENABLE_TUVX=ON \
+                              -D MUSICA_ENABLE_PYTHON_LIBRARY=ON
+
```
boulderdaze, line 164: My open PR removes duplicate mod files in the include path and ensures that those mod files are only built and installed when `MUSICA_BUILD_FORTRAN_INTERFACE` is enabled. I’m curious if you need this option enabled

boulderdaze (changes requested):
Looks good, I just have a question.

.github/workflows/windows.yml
boulderdaze: Maybe it's just the installation, not the build. I’ll check the build include paths

boulderdaze: These options still build the mod files, but not the `lmusica` or `lmusica-fortran`. I want to fix the build include so that mod files are only created when the `MUSICA_BUILD_FORTRAN_INTERFACE` is enabled. Does this make sense?

boulderdaze: I’m a bit confused because TUV-x is enabled, so it makes sense to have the TUV-x mod files, but the TUV-x C++ header files aren’t there. I think it should be the other way around. I've changed my PR to a draft and will work on it. But could you add `MUSICA_BUILD_FORTRAN_INTERFACE` option since it tests the Fortran code?

These are not found in include for the specified options 
```CPP
├── grid.hpp
├── grid_map.hpp
├── profile.hpp
├── profile_map.hpp
├── radiator.hpp
├── radiator_map.hpp
└── tuvx.hpp
```

K20shores: good catch

boulderdaze (approved):
Thank you!

PR #254, opened 2025-02-19
Add map matching options

mattldawson:
Adds the ability to create index mappings that match every source-target pair in the configuration file, or only those where the source and target labels are both present in the source and target arrays.

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -217,7 +217,6 @@ namespace musica
         }
       } else if (!IsSuccess(*error))
       {
-        DeleteIndexMappings(&index_mappings);
         return index_mappings;
       }
       std::size_t target_index = FindMappingIndex(target, target_name.c_str(), error);
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -185,36 +185,70 @@ namespace musica
   }
 
   IndexMappings
-  CreateIndexMappings(const Configuration configuration, const Mappings source, const Mappings target, Error* error)
+  CreateIndexMappings(const Configuration configuration, const IndexMappingOptions map_options, const Mappings source, const Mappings target, Error* error)
   {
     DeleteError(error);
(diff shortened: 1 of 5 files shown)


src/util.cpp
```diff
@@ -185,36 +185,70 @@ namespace musica
...
+        {
+          return index_mappings;
+        }
+        else
+        {
+          DeleteError(error);
+          *error = NoError();
+          continue;
+        }
+      } else if (!IsSuccess(*error))
```
boulderdaze, line 218: When an error occurs and the mapping option is set to `MapAll`, it seems that the mapping fails. I think this could follow the same logic as when `IsSuccess` fails, where the mappings are cleaned up and an empty object is returned. I'm curious why the mappings are cleaned up in one case, but not in the other.
```cpp
{
  DeleteIndexMappings(&index_mappings);
  return index_mappings;
}
```
It also looks like in line 238, the call to `DeleteIndexMappings(&index_mappings)` is missing, which is called in line 220 for the same conditional.

boulderdaze (changes requested):
Looks good! I have a question, but everything looks good overall.

src/util.cpp
mattldawson: good point, we should be consistent. I will update this to delete the mappings object (or possibly just not allocate it until the end of the function)

mattldawson: Actually, after looking at it again, it looks like the allocation is not happening until the end, so these return statements should just be returning structs with pointers to `nullptr`. Let me know what you think.

boulderdaze: I see. Does it mean this call `DeleteIndexMappings(&index_mappings)` in line 220 is also unnecessary?

mattldawson: ah, yes - good point. I removed it.

boulderdaze (approved):

PR #257, opened 2025-02-20
adding mechanism configuration parser dependency

### LINKED ISSUE #255: Add Mechanism Configuration as a dependency
[Mechanism Configuration](https://github.com/open-atmos/MechanismConfiguration)

K20shores:
Closes #255 

This is also in support of #253. It also needs the PR from [open atmos](https://github.com/open-atmos/MechanismConfiguration/pull/58).


mattldawson (approved):

cmake/dependencies.cmake
```diff
@@ -127,3 +127,17 @@ endif()
...
+set_git_default(MECH_CONFIG_GIT_REPOSITORY https://github.com/open-atmos/MechanismConfiguration.git)
+set_git_default(MECH_CONFIG_GIT_TAG 0835527)
+
+FetchContent_Declare(mechanism_configuration
+    GIT_REPOSITORY ${MECH_CONFIG_GIT_REPOSITORY}
+    GIT_TAG ${MECH_CONFIG_GIT_TAG}
+    GIT_PROGRESS NOT ${FETCHCONTENT_QUIET}
+)
+
+FetchContent_MakeAvailable(mechanism_configuration)
```
boulderdaze, line 143: I think it would be nice to move this fetch content after fetching the netcdf and yaml, as it doesn’t require condition checks like those libraries do.

boulderdaze (approved):
Looks good. I have one suggestion, but it should't prevent your PR from being merged

cmake/dependencies.cmake
K20shores: Good catch, moved

PR #266, opened 2025-03-05
Provide interface micm state 217

### LINKED ISSUE #217: Provide an interface to the micm State

montythind:
This PR includes refactoring the code and creating an interface to pass State directly from Python to micm directly. It will remove temperature, pressure, air_density, ordered_concentrations, ordered_rate_constants

closes #217


K20shores (changes requested):
A few comments, we also need some tests for the C API around the state please

boulderdaze:
Hi @montythind, could you resolve the merge conflicts? It seems like the tests didn't run

K20shores (approved):

boulderdaze (changes requested):

cmake/dependencies.cmake
```diff
@@ -33,7 +33,7 @@ FetchContent_MakeAvailable(yaml-cpp)
 ################################################################################
 # Mechanism Configuration
 
-set_git_default(MECH_CONFIG_GIT_REPOSITORY https://github.com/open-atmos/MechanismConfiguration.git)
+set_git_default(MECH_CONFIG_GIT_REPOSITORY https://github.com/NCAR/MechanismConfiguration.git)
```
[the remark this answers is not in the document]
K20shores: @boulderdaze we moved the mech config repository from open-atmos to the NCAR org. We now own it entirely

boulderdaze (changes requested):
I like the approach of creating a wrapper around the state. It involves a lot of changes across various locations, and I think you did a great job. One thing I wanted to mention is that the documentation is incorrect in several places — including function / type descriptions, parameters, copyright years, indentation. I’ve left comments on some of them, but not all, and I’d like to do a re-review once the documentation is updated.

cmake/dependencies.cmake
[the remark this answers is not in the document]
boulderdaze: I see, thank you

src/micm/micm.cpp
```diff
@@ -407,65 +321,104 @@ namespace musica
...
+            *solver_stats = SolverResultStats(
+                result.stats_.function_calls_,
+                result.stats_.jacobian_updates_,
+                result.stats_.number_of_steps_,
+                result.stats_.accepted_,
+                result.stats_.rejected_,
+                result.stats_.decompositions_,
+                result.stats_.solves_,
+                result.final_time_);
+        };
```
boulderdaze, line 347: Why is this lambda function defined inside the try block?

boulderdaze (changes requested):
Thank you for addressing my code review comments. I still have a couple more, and I hope it won’t take too much of your time to address them.

src/micm/micm.cpp
montythind: I was trying to follow the existing template for this. But i think some of the method solver->CalculateRateConstants and solver->solve throw exceptions if they fail.

boulderdaze: Okay, it's a function definition, so I was curious why it's inside a try block. But it's fine either way.

boulderdaze (approved):
Thank you for addressing the review comments. In the functions that retrieve species orderings and rate constant orderings, instead of the `MUSICA_ERROR_CODE_SOLVER_TYPE_NOT_FOUND` error, I was expecting an error indicates that the state type not being found, but I’ve made lots of requests, so feel free to implement this in your next working PR. Thank you.

include/musica/micm/micm.hpp
```diff
@@ -116,42 +117,71 @@ namespace musica
...
-        double *temperature,
-        double *pressure,
-        double *air_density,
-        double *concentrations,
-        double *custom_rate_parameters,
-        String *solver_state,
-        SolverResultStats *solver_stats,
-        Error *error);
+      MICM *micm,
+      musica::State *state_wrapper,
```
mattldawson, line 127: My suggestion would be to just call this `state` instead of `state_wrapper`, just because the users shouldn't have to be aware that this is a wrapper of our MICM State class. But, no strong objections if the group prefers this naming instead.

(similar comment for the other functions in this header)

mattldawson (approved):
Looks great! Just one minor suggestion, but nothing that should hold up merging.

include/musica/micm/micm.hpp
boulderdaze: I like this idea

montythind: done

K20shores (approved):

PR #270, opened 2025-03-18
258 add documentation for reading configurations

### LINKED ISSUE #258: Add documentation for reading configurations

K20shores:
Closes #258 

I made a few minor updates to the docs to better lay out the user guide. However, Chapter 2 of the tutorial already talks about configurations. I figure we can wait on adding things for version 1 of the configuration once we are sure we have settled on the format. Also, it's possible we will soon have functionality in the mechanism configuration to export formats after making them in python, which could easy writing these tutorials


docs/CMakeLists.txt
```diff
@@ -13,8 +13,6 @@ add_custom_command(
   WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
   MAIN_DEPENDENCY ${SPHINX_SOURCE}/conf.py
   COMMENT "Generating documentation with Sphinx"
-  # run this command each time. This way we don't have to keep track of each individual rst file
-  BuildDocs
```
mattldawson, line 17: is this no longer needed?

mattldawson (approved):
looks good! just one question

docs/CMakeLists.txt
K20shores: This is apparently incorrect and in the middle of correcting this, I got distracted. It's fixed now

K20shores: Oh, actually it isn't needed. It was meant to be a label to use to build, like make BuildDcos, but we already have one, make docs

PR #283, opened 2025-04-02
Correct python builds

K20shores:
This PR reworks how we build the python library.

- now we link to the musica library; we no longer have to list each file that we need to build
  - we have to explicity set the rpath for musica
- we also link to mechanism configuration, I had to set the [rpath for that library](https://github.com/NCAR/MechanismConfiguration/pull/77)
- adds a build argument that allows us to yaml-cpp which has a cmake version that isn't supported by cmake anymore
- Python CI builds are now enables for all PRs to ensure we can always do a release cleanly


cmake/dependencies.cmake
```diff
@@ -19,22 +19,11 @@ if (MUSICA_BUILD_FORTRAN_INTERFACE)
...
-  GIT_PROGRESS NOT
-  ${FETCHCONTENT_QUIET})
-FetchContent_MakeAvailable(yaml-cpp)
-
 ################################################################################
 # Mechanism Configuration
 
 set_git_default(MECH_CONFIG_GIT_REPOSITORY https://github.com/NCAR/MechanismConfiguration.git)
-set_git_default(MECH_CONFIG_GIT_TAG v0.1.0)
+set_git_default(MECH_CONFIG_GIT_TAG rpath)
```
mattldawson, line 26: just a note that this will need updated when the other PR is merged

mattldawson (approved):
nice!

PR #295, opened 2025-04-14
Refactor errors usage, micm API

K20shores:
This PR refactors the C API to MICM so that the `Error` object is only used in one place. This should reduce the number of memory errors we encounter since there is exactly one place for the API to delete and create errors.

I also updated the usage of errors in micm so that we start using error categories and system errors throughout the musica codebase. This allows the new error handling to properly provide errors back to fortran.

I removed the hard-coded errors in `util.F90`, which are redundant since these can be included with `#include "musica/error.hpp"` in fortran.

I moved the C interface into its own file so that the C++ interface is separate.

I removed the `Error` object from all of the `MICM` C++ interfaces and refactored the `MICM` class so that its constructor creates the solver in a cleaner way, information that is already stored on the underlying `micm::Solver` isn't stored twice, and added a visitor struct to clean up the `Solve` interface. This is in preparation for the Cuda solver, which will need some extra steps before and after solving and will be far easier to do with the visitor struct than the old `constexpr` interface

Depends on https://github.com/NCAR/micm/pull/758


mattldawson (approved):
nice!

include/musica/micm/micm.hpp
```diff
@@ -27,310 +20,136 @@
 #include <utility>
 #include <vector>
 
-#ifndef MICM_VECTOR_MATRIX_SIZE
-  #define MICM_VECTOR_MATRIX_SIZE 4
-#endif
-
-namespace musica
+enum class MusicaErrc
```
boulderdaze, line 23: Since the class `MusicaErrorCategory` is defined with a fully descriptive name, I think `enum class MusicaErrc` could also follow suit—perhaps something like `MusicaErrorCode` or another fully descriptive name

boulderdaze (approved):

PR #312, opened 2025-05-01
In-code mechanism creation

### LINKED ISSUE #144: Create a musica python library
The wrapper provides all of the musica functions in python. However, we can easily check all of the preconditions (like passing `None` to an argument that expects a double will still work but there is no way to detect this in C). 

Write a small musica python package which wraps the musica C library that is strictly meant to do type checking before passing data to musica

## Ideas
- [moab](https://bitbucket.org/fathomteam/moab/src/master/pymoab/) could be a template for us

mattldawson:
Adds the ability to create mechanisms in python using the latest version of the MechanismConfiguration code. Also creates a set of wrapper classes for the python API with the ability set state data from dictionaries for any of the underlying MICM matrix types validate arguments.

closes #144


musica/test/test_analytical.py
```diff
@@ -0,0 +1,344 @@
+import unittest
+import numpy as np
+import musica
+import random
+import mechanism_configuration as mc
+from _musica._core import _is_cuda_available
+
+
+def TestSingleGridCell(self, solver, state, time_step, places=5):
+    num_grid_cells = 1
```
boulderdaze, line 10: `num_grid_cells = 1` isnt' used. Should we remove it?

musica/test/test_chapman.py
```diff
@@ -0,0 +1,119 @@
...
+    O2 = mc.Species(name = "O2", tracer_type = "CONSTANT")
+    O = mc.Species(name = "O", other_properties = {"__absolute_tolerance": "1e-12"})
+    O1D = mc.Species(name = "O1D", other_properties = {"__absolute_tolerance": "1e-12"})
+    O3 = mc.Species(
+         name = "O3",
+         molecular_weight_kg_mol = 0.048,
+         other_properties = {
+              "__absolute_tolerance": "1e-12",
+              "__long_name": "ozone",
+              "__atmos": "3",
```
boulderdaze, line 50: Just curious, what this property `"__atmos": "3"` is for?

docker/Dockerfile.python-tests
```diff
@@ -0,0 +1,40 @@
...
+
+RUN cd /build/_deps/mechanism_configuration-src \
+    && python3 -m pip install --upgrade wheel setuptools pytest \
+    && pip3 install --verbose .[test]
+
+RUN cd /musica \
+    && pip3 install --verbose .[test]
+
+WORKDIR /musica
+CMD ["python3", "-m", "unittest", "discover", "-v", "-s", "musica/test"]
```
boulderdaze, line 40: It doesn't seem that this dockerfile is tested via github actions, but `python-test.yml` looks like covering the tests based on the identical-looking script. Would this be for local testing?

musica/test/test_analytical.py
mattldawson: removed!

musica/test/test_chapman.py
mattldawson: I'm not sure (I just renamed the folder this test was in), but I think it's just testing that any random key-value pair can be added to `other_properties`

docker/Dockerfile.python-tests
mattldawson: Yeah, I used this for debugging the python API locally, but I figured it wasn't worth adding to the actions since the `python-test.yml` is already doing these tests (including for different OS/Python versions)

sjsprecious (approved):

PR #315, opened 2025-05-02
Update API to use latest version of MICM

mattldawson:
Updates the C, Fortran, an Python APIs to use MICM's new solver and state constructors that move ownership of the number of grid cells to the state.

This will allow CAM-SIMA to construct a MICM solver before knowing the number of grid cells each MPI rank will solve. Also, it will allow python users to manage collections of grid cells that aren't easily divisible by the vector length when using MICM's vector-ordered matrices.


sjsprecious (approved):
Thanks @mattldawson for updating the API.

boulderdaze (approved):
Looks good! I just have a single minor request

PR #319, opened 2025-05-03
Update python state

mattldawson:
Updates the Python API to set state data in-place and avoid having to specify all species/parameters when updating the state.

### DIFF
--- a/musica/wrapper.cpp
+++ b/musica/wrapper.cpp
@@ -9,14 +9,20 @@
 
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
+#include <pybind11/stl_bind.h>
 
 namespace py = pybind11;
 namespace v1 = mechanism_configuration::v1::types;
+
+PYBIND11_MAKE_OPAQUE(std::vector<double>)
+PYBIND11_MAKE_OPAQUE(std::vector<micm::Conditions>)
 
 // Wraps micm.cpp
 PYBIND11_MODULE(_musica, m)
 {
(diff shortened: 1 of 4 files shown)


sjsprecious (approved):

musica/wrapper.cpp
```diff
@@ -9,14 +9,20 @@
 
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
+#include <pybind11/stl_bind.h>
 
 namespace py = pybind11;
 namespace v1 = mechanism_configuration::v1::types;
 
+PYBIND11_MAKE_OPAQUE(std::vector<double>)
+PYBIND11_MAKE_OPAQUE(std::vector<micm::Conditions>)
```
K20shores, line 18: I think this means these are copied, according to the [docs](https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html):

> The major downside of these implicit conversions is that containers must be converted (i.e. copied) on every Python->C++ and C++->Python transition, which can have implications on the program semantics and performance. Please read the next sections for more details and alternative approaches that avoid this.

K20shores: Or maybe I'm misunderstanding. It seems that `PYBIND11_MAKE_OPAQUE` allows objects from C++ to be passed by reference. So never mind, I think

mattldawson: Yeah, this was actually the only way I could get the state arrays to be passed by reference. Without this, they were copies and when you tried to modify them in Python, you just modified the copies. Now, it actually modifies the state arrays directly.

boulderdaze (approved):
Looks good!

PR #324, opened 2025-05-06
Allow ints to be passed to set conditions

mattldawson:
Updates the set_conditions function in the Python API to allow setting temperature, pressure, or air density with int values.

### DIFF
--- a/musica/types.py
+++ b/musica/types.py
@@ -67,9 +67,9 @@ class Conditions(_Conditions):
 
     def __init__(
         self,
-        temperature: Optional[Union[float | int]] = None,
-        pressure: Optional[Union[float | int]] = None,
-        air_density: Optional[Union[float | int]] = None,
+        temperature: Optional[Union[float, int]] = None,
+        pressure: Optional[Union[float, int]] = None,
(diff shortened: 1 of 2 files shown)


musica/types.py
```diff
@@ -195,15 +192,15 @@ def set_conditions(self,
         air_densities : Optional[Union[float, List[float]]]
             Air density in mol m-3. If not provided, it will be calculated from the Ideal Gas Law.
         """
-        if isinstance(temperatures, float):
+        if isinstance(temperatures, float) or isinstance(temperatures, int):
```
boulderdaze, line 195: why does it allow `int` for temperature, pressure and density?

boulderdaze (approved):
Looks good! I just have one question

musica/types.py
mattldawson: It's for if someone would right something like:
```
state.set_conditions(temperature=298, pressure=101325)
```
Prior to this change, this would throw an error. (There is already similar logic for ints in concentrations and rate parameters)

boulderdaze: ah, I see, thanks!

K20shores (approved):

PR #327, opened 2025-05-07
Disable building tests for Mechanism Configuration

boulderdaze:
Disable building tests for Mechanism Configuration


K20shores:
@boulderdaze the python tests fail for a reason that we are in the middle of fixing. I think a new release of mechanism configuration will fix it, so don't worry about those failing tests

K20shores (approved):

PR #328, opened 2025-05-08
moving mechaning configuration python module to musica

K20shores:
I ran into an issue with ABI compatibility for some reason. The issue is that both musica and mechanism configuration need the mechanism configuration static library. I filed a [discussion](https://github.com/pybind/pybind11/discussions/5645), but honestly the easiest answer is to move mechanism configuration python code to musica. Only musica uses it anyways, so may as well put it here.


mattldawson (approved):

boulderdaze:
I read the issue that you created in pybind repo and had a few questions. As for the descriptions about two separate builds of the Mechanism Configuration library being linked into the same Python process — could it be related to how we've set up the CMake? From what I understand, we include mechanism configuration as a target when exporting musica library, under the assumption that this is part of musica. I'm starting to wonder if we should avoid packaging mechanism configuration and instead require  users should build/install it separately (adding `find_library(mechanism_configuration)` in the exporting configuration) similar to how we require users to build/install net-cdf library separately.
Also, if a struct of, for example, `HenrysLaw` changes, do we now have to update in both mech config and musica? If so, I think that is something we have to consider.

K20shores:
@boulderdaze, yeah, requiring an install ahead of time seems like it would fix this. Part of the issue is that C++ depends on the library, and so does the Python `.so` file. To make it work, we would need to install a mechanism configuration system-wide and then use `find_package` in both musica *and* the mechanism configuration python library. But that means that for someone to install the mechanism configuration, they would have to install it twice, once for the system and again for Python. That seems like a lot of work, and there has to be a better solution, but I'm not sure what it is.

That said, we are the only people using mechanism configuration. If something changes, updating the mechanism configuration tag will cause a build error, and we will fix that when we encounter it. 

I would love to figure out a better way to do this, but I also don't want to spend that much more time on this.

boulderdaze:
I see your point, but this approach makes maintenance more difficult since it requires updating the code in two places. But yes, if we don't want to put more time in this, we can revisit this later when we have the bandwidth. 
By the way, is  `config/examples` used only for pytest? If so, should we consider moving it under the to `test` directory? 

> @boulderdaze, yeah, requiring an install ahead of time seems like it would fix this. Part of the issue is that C++ depends on the library, and so does the Python `.so` file. To make it work, we would need to install a mechanism configuration system-wide and then use `find_package` in both musica _and_ the mechanism configuration python library. But that means that for someone to install the mechanism configuration, they would have to install it twice, once for the system and again for Python. That seems like a lot of work, and there has to be a better solution, but I'm not sure what it is.
> 
> That said, we are the only people using mechanism configuration. If something changes, updating the mechanism configuration tag will cause a build error, and we will fix that when we encounter it.
> 
> I would love to figure out a better way to do this, but I also don't want to spend that much more time on this.

K20shores:
> I see your point, but this approach makes maintenance more difficult since it requires updating the code in two places. But yes, if we don't want to put more time in this, we can revisit this later when we have the bandwidth. By the way, is `config/examples` used only for pytest? If so, should we consider moving it under the to `test` directory?
> 
> > @boulderdaze, yeah, requiring an install ahead of time seems like it would fix this. Part of the issue is that C++ depends on the library, and so does the Python `.so` file. To make it work, we would need to install a mechanism configuration system-wide and then use `find_package` in both musica _and_ the mechanism configuration python library. But that means that for someone to install the mechanism configuration, they would have to install it twice, once for the system and again for Python. That seems like a lot of work, and there has to be a better solution, but I'm not sure what it is.
> > That said, we are the only people using mechanism configuration. If something changes, updating the mechanism configuration tag will cause a build error, and we will fix that when we encounter it.
> > I would love to figure out a better way to do this, but I also don't want to spend that much more time on this.

Yes, it's only for the tests. I'll move it there

boulderdaze (approved):

PR #331, opened 2025-05-08
Add pkg-config to install

mattldawson:
Adds a `pkg-config` file to the install lib directory for easier inclusion in builds.

Removes a few tests of the MUSICA Fortran API that were directly using TUV-x modules, which isn't supported when compiling the MUSICA and MUSICA-Fortran libraries with different compilers.


K20shores (approved):

CMakeLists.txt
```diff
@@ -58,6 +58,23 @@ elseif(${CMAKE_Fortran_COMPILER_ID} MATCHES "PGI")
...
+    set(CMAKE_CXX_STANDARD_LIBRARIES "-lstdc++")
+  elseif(${CMAKE_CXX_COMPILER_ID} MATCHES "IBM")
+    set(CMAKE_CXX_STANDARD_LIBRARIES "-lstdc++")
+  elseif(${CMAKE_CXX_COMPILER_ID} MATCHES "NVHPC")
+    set(CMAKE_CXX_STANDARD_LIBRARIES "-lstdc++")
+  elseif(${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
+    set(CMAKE_CXX_STANDARD_LIBRARIES "-lstdc++") # use GNU standard library
+  else()
+    set(CMAKE_CXX_STANDARD_LIBRARIES "-lstdc++") # default to most common name
+  endif()
```
boulderdaze, line 75: Every compiler sets `CMAKE_CXX_STANDARD_LIBRARIES` variable as `-lstdc++` even in the default case, which makes conditional checks to have no effect. I think we should remove the conditional checks and  set it as a general variable. 
The fact that compiler is being checked makes me think that different compilers may use different names for C++ std library. Would it be that case?

boulderdaze (commented):
Looks good to me! I just have one question.

CMakeLists.txt
mattldawson: From what I read it seems like most compilers use `libstdc++` as the name, except Clang, which uses `libc++`. But when I went to use `libc++` with the Clang tests, it couldn't find the library and it turns out Clang by default uses the GNU standard library (`libstdc++`).

The reason I left the conditional blocks was just as a reminder that if we try this with a new compiler, there's a chance the standard library name may be different. But, I could also remove compiler-specific logic if that's clearer.

boulderdaze: Okay, I see your point. I don't request any changes then.

PR #352, opened 2025-05-19
Force RPATH to look in site packages for linux wheel

K20shores:
In #343 I disabled the GPU build of the python wheels on linux because the cuda files were being bundled with our wheel file. That made the wheel file too large to upload to pypi.

Nvidia publishes wheel files on their own index that contain cuda shared object libraries. You can view all of the packages they have available [here](https://pypi.nvidia.com/).  They also provide instructions for how to install those here: https://docs.nvidia.com/cuda/cuda-installation-guide-linux/#prerequisites

This PR forces the RPATH to look in the python site packages folder, similar to what [pytorch](https://github.com/pytorch/pytorch/blob/7a46f4bde0240521321fa024af1e38a457d1b0cd/.ci/manywheel/build_cuda.sh#L204) does. To do this, I have to open the wheel file, directly write the RPATH, and I also delete the cuda so files that were packages. In addition, I added an optional dependency set that will install the cuda runtime python packages that can be tested with `pip install .[gpu]` on derecho. This means that for someone to get GPU support from pypi, it would only be `pip install musica[gpu]` and they shouldn't have to compile by hand.

I also updated the pyproject.toml cmake arguments. Apparently it's beter to use `[cmake.define](https://scikit-build-core.readthedocs.io/en/latest/configuration/index.html#configuring-cmake-arguments-and-defines)` and to [override](https://scikit-build-core.readthedocs.io/en/latest/configuration/overrides.html#inheriting-for-tables-and-arrays) those for different build configurations.


mattldawson (approved):

CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.21)
 
 # must be on the same line so that pyproject.toml can correctly identify the version
-project(musica-distribution VERSION 0.11.1)
+project(musica-distribution VERSION 0.11.1.0)
```
boulderdaze, line 4: Are we using the tweak in versioning now?

boulderdaze (approved):
Looks good!

CMakeLists.txt
K20shores: only for me to test out the published build

PR #385, opened 2025-06-25
Contributors, Authors, and Citations

K20shores:
This creates a dedicated Contributors guide, a page for authors, and adds a `.zenodo.json` which allows us to more explicitly control how we appear on zenodo. Additionally, I learned that we can specify contributor types in the zenodo file. These types are listed [here](https://datacite-metadata-schema.readthedocs.io/en/4.6/appendices/appendix-1/contributorType/).


.zenodo.json
```diff
@@ -0,0 +1,75 @@
...
+    },
+    {
+      "name": "Ward Fisher",
+      "type": "Other"
+    },
+    {
+      "name": "Jian Sun",
+      "type": "ProjectMember"
+    }
+  ],
```
mattldawson, line 59: I think we could put Jian in with the creators, what do you think? Also, we should add Qina Tan here, since she wrote the majority of the CUDA solver code

mattldawson (approved):
I like it! Just one suggestion, but I think it's fine to merge

.zenodo.json
K20shores: Fine with me

sjsprecious (approved):
LGTM.

PR #391, opened 2025-06-26
Fix some of the warnings which were suppressed in silence_cmake file

montythind:
This PR includes fixing the warnings which were suppressed in silence_cmake file. 
Some of the warnings supressed are:  GetSpeciesOrdering, GetUserDefinedRateParametersOrdering, GetSpeciesPropertyString, MicmVersion, LoadConfigurationFromFile etc.


K20shores (approved):

boulderdaze:
I understand that Copilot generates a summary for the PR, but having an unclear title with no description can be confusing for reviewers. It doesn't effectively communicate the changes this PR introduces to other developers.

src/util.cpp
```diff
@@ -184,22 +178,23 @@ namespace musica
...
       const Mappings target,
+      IndexMappings *indexMappings,
       Error* error)
   {
     DeleteError(error);
     std::size_t size = configuration.data_->size();
     std::vector<IndexMapping> mappings;
-    IndexMappings index_mappings;
-    index_mappings.size_ = 0;
+    //IndexMappings index_mappings;
```
mattldawson, line 192:
```suggestion
```

mattldawson (changes requested):
Looks good! Just a few minor suggestions

boulderdaze (commented):
Thank you for addressing the comment. It looks like the functions have been changed to subroutines, and their return values are now being passed as parameters instead. Could you explain the reasoning behind this change, how it is related to dealing with the warnings?
Also, the copilot suggestions look good to me. Have you had a chance to review them?

montythind:
The change is necessary because returning a struct with pointers (like Mappings, output) by value is not compatible with C or Fortran. It causes compiler warnings or undefined behavior when crossing language boundaries. By changing as an example  to void GetSpeciesOrdering(musica::State* state, Mappings* output, Error* error), make it compatible with extern 'c' and avoids warnings.

mattldawson (changes requested):
Getting closer! I think just one last thing: our style is to have the function arguments be `snake_case`

PR #406, opened 2025-07-07
moving backend import logic to its own file

### LINKED ISSUE #370: Python files
After #362 went in, an automatic PR was made to correct some python file format #368. I rejected that because it would have caused a circular import:

```
>>> import musica
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/kshores/Documents/musica/musica/__init__.py", line 1, in <module>
    from .types import MICM, SolverType, State, Conditions
  File "/Users/kshores/Documents/musica/musica/types.py", line 9, in <module>
    from musica import (
ImportError: cannot import name '_Conditions' from partially initialized module 'musica' (most likely due to a circular import) (/Users/kshores/Documents/musica/musica/__init__.py)
```

**Acceptance criteria**
- No circular imports exist
- Autoformat PRs can be merged without worrying about breaking the python fiels

**Ideas**
- Rewrite the `__init__.py` and other python files so that no circular import exists
- Maybe figure out how to remove the disgusting things we had to add in #362 that import symbols by name from each musica backend
- Possibly just flag this part of the code in a way that prevents it from being reformatted.

K20shores:
The `__init__.py` file would get a circular import if the imports in `__init__.py` were re-ordered. This is because of the conditional import logic we need to do of the pybind11 musica targets for cpu vs gpu. This, I think, should fix that circular import and allow our python format actions to be turned on again

Closes #370

### DIFF
--- a/musica/__init__.py
+++ b/musica/__init__.py
@@ -1,51 +1,11 @@
-import importlib.util
+"""
+MUSICA: A Python library for atmospheric chemistry simulations.
+"""
 
-
-def _safe_find_spec(name):
-    try:
-        return importlib.util.find_spec(name)
-    except ModuleNotFoundError:
-        return None
-
-
-def _gpu_deps_installed():
-    return (
-        _safe_find_spec("nvidia.cublas") is not None or
-        _safe_find_spec("nvidia_cuda_runtime") is not None or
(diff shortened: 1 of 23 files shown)


boulderdaze (approved):
Looks good!

Issue #400, opened 2025-06-30
Create a TUVX python wrapper


K20shores:
TUVx can be fully configured from a config file. Create a wrapper that takes a config and returns photos rates

**Acceptance criteria**
- A configuration file can be used to create TUVX
- Tests exist to show basic functionality to retrieve and label photo rates
- The wrapper builds on at least macOS and Linux because it's easy to get a fortran compiler there

**Ideas**
- Model it after the musica wrapper
- Extend the TUVX class as it is

PR #409, opened 2025-07-08
Add tuvx into the python build for linux and macos

K20shores:
This is the second of 3 PRs to address #400 

The first PR I tried to make was #393, but that ended up being too large and I split it into 3. The first was #406 

This PR adds TUVX into the python build. To do that, I had to use a newer linux image when building the python wheels to get netcdf into the build easily. At this point, we only support macOS and Linux, but it may be possible to support windows; I don't know how to get a fortran compiler on windows with cibibuildhweel and felt like leaving that issue for later.

I created a simple python tuvx class as a place holder and am able to return the tuvx version in python


mattldawson (approved):
nice!

PR #537, opened 2025-07-31
Combine formatting actions

K20shores:
This combines the clang and autopep8 format actions into one action so that we don't have 2 formatting prs to merge in each time


boulderdaze (approved):
Like it!

PR #569, opened 2025-08-12
Correct debug build of python library

### LINKED ISSUE #558: Correct debug build of musica-python for carma
On macOS, at least, building the musica python library in debug causes the python code to stall in the tests. Investigate this and fix

**Acceptance criteria**
- Debug build of python library on macOS has passing and finishing tests

**Ideas**
- Remove the suppression in the valigrind file and run memcheck on the python tests after building the project in debug mode in a docker container. While this is on linux, this may identify some memory issue that could be causing this to fail

K20shores:
- Fixed gcc 13, 14 builds by using [maxim-lobanov/setup-xcode](https://github.com/maxim-lobanov/setup-xcode) and seleecting Xcode 16.4 as recommended in [this issue](https://github.com/actions/runner-images/issues/12745)
- Update the intel and nvidia docker tags to use a newer cmake version
- Add try/catch blocks around vector of double conversion and doing so by hand. WIthout this, for some reason we were getting a bad alloc error and it crashed the python process

closes #558

### DIFF
--- a/docker/Dockerfile.fortran-intel
+++ b/docker/Dockerfile.fortran-intel
@@ -1,5 +1,4 @@
 # versions and sizes from here: https://hub.docker.com/r/intel/oneapi-hpckit/tags
-# FROM intel/oneapi-hpckit:2024.0.1-devel-ubuntu22.04
 FROM intel/oneapi-hpckit:2025.2.0-0-devel-ubuntu24.04
 
 


docker/Dockerfile.fortran-intel
```diff
@@ -1,19 +1,19 @@
...
+FROM intel/oneapi-hpckit:2025.2.0-0-devel-ubuntu24.04
+
 
 # Based off of this: https://dgpu-docs.intel.com/driver/installation.html#repository-public-key-used-for-package-and-repository-signing
 # however those docs (at the time of this writing are incorrect) and this is the correct url
 # community.intel.com/t5/Intel-oneAPI-Math-Kernel-Library/Cannot-update-any-packages-using-intel-oneapi-hpckit-latest/m-p/1603192#M36169
-RUN wget -qO - https://repositories.intel.com/gpu/intel-graphics.key | \
-  gpg --yes --dearmor --output /usr/share/keyrings/intel-graphics-archive-keyring.gpg
+# RUN wget -qO - https://repositories.intel.com/gpu/intel-graphics.key | \
+#   gpg --yes --dearmor --output /usr/share/keyrings/intel-graphics-archive-keyring.gpg
```
boulderdaze, line 10: Do we no long need to download a GPG key anymore? If so should we remove those lines?

boulderdaze (commented):
Looks good! Just a couple requests.

docker/Dockerfile.fortran-nvhpc
```diff
@@ -4,14 +4,14 @@
 #
 # Container versions, and sizes, can be found at https://catalog.ngc.nvidia.com/orgs/nvidia/containers/nvhpc/tags
 #
-FROM nvcr.io/nvidia/nvhpc:23.7-devel-cuda12.2-ubuntu22.04
+# FROM nvcr.io/nvidia/nvhpc:23.7-devel-cuda12.2-ubuntu22.04
+FROM nvcr.io/nvidia/nvhpc:25.7-devel-cuda12.9-ubuntu24.04
```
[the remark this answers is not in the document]
K20shores: Good point, yes I'll remove it

docker/Dockerfile.fortran-intel
K20shores: We aren't but I remember it took me a *long* time to find this gpg thing the first time so I'd like to leave it in case the intel file breaks again and we need this

boulderdaze: okay, no problem!

boulderdaze (approved):

PR #585, opened 2025-08-15
Building windows python wheels with fortran support

### LINKED ISSUE #590: Include fortran builds on windows

K20shores:
Closes #590

- The python tests actions now builds wheels for all operating system, and then uses these wheels to run the tests. 
  - This is a much better because it more closely follow what a user might get from pypi
- Added msys2 to provide the mingw build tools for windows which includes gfortran
  - [using msys2 in ci](https://www.msys2.org/docs/ci/)
  - [arm support](https://www.msys2.org/wiki/arm64/)
    - Unused but leaving the composite action for use later

**Attempts**
- Attempted support for windows arm
  - LLVM's flang on windows arm for TUVx and CARMA
  - Abandoned since there were symbol visibility issues 
- Attempted and then abandoned rtools to provide mingw on windows which provides gfortran
  - Failed because we need to install other packages, and pacman intermittently failed with no easy resolution in sight

**Other information**
At first I tried to use rtools, like in scipy, so I heavily relied on what scipy does to support fortran on windows.
- [this workflow](https://github.com/scipy/scipy/blob/main/.github/windows_arm64_steps/action.yml) sets up Flang and pkg-config on arm windows. I made a similar one for amd based windows
- Their [testing action](https://github.com/scipy/scipy/blob/main/.github/workflows/wheels.yml)
- Their [windows action](https://github.com/scipy/scipy/blob/main/.github/workflows/windows.yml) which helped me figure out how to configure windows to include a build that needs fortran. This also showed me that we should be building wheels, installing those wheels, and testing our files that way
- Their [repair](https://github.com/scipy/scipy/blob/main/tools/wheels/repair_windows.sh) bash script for wheel files

### DIFF
--- a/.github/windows_arm64_steps/action.yml
+++ b/.github/windows_arm64_steps/action.yml
@@ -4,7 +4,7 @@ description: Setup MSYS2 ARM64 for CI
 # this is currently not used because there was a compiler issue
 # where the visibilty flags set by pybind11 caused build failures
 # on Windows ARM64 with Clang. Leaving this action here for future use
-# becuase it was difficult for me to figure out what packages were needed
+# because it was difficult for me to figure out what packages were needed
 
 runs:
   using: "composite"
--- a/.github/workflows/python-wheels.yml
+++ b/.github/workflows/python-wheels.yml
@@ -76,7 +76,7 @@ jobs:
       if: matrix.os == 'windows-latest'
       uses: ./.github/windows_amd64_steps
 
-    # the if implicity uses the fact that arm on windows is not running on windows-latest
+    # the if implicitly uses the fact that arm on windows is not running on windows-latest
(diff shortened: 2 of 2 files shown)


python/musica/tools/prepare_build_environment_windows.bat
```diff
@@ -0,0 +1,66 @@
+<<<<<<< HEAD:python/musica/tools/prepare_build_environment_windows.bat
+# PowerShell script to set up Windows build environment
```
boulderdaze: Could you remove these conflict marks?

K20shores: I  have deleted this file like 8 times and it keeps coming back. It's not supposed to be included

PR #595, opened 2025-09-04
Add dose rates to TUV-x Python output

mattldawson:
This PR adds dose rates as an output to the TUV-x Python API.

Also:
- replaces stubbed out functions for rate labels with existing functions in the TUV-x C/C++ API.
- adds some additional testing
- removes TUV-x diagnostic output that is only needed for the internal TUV-x regression testing
- allows creation of a TUV-x calculator based on a JSON/YAML string (in addition to the existing ability to read the configuration data from a file)


include/musica/tuvx/tuvx.hpp
```diff
@@ -88,19 +100,25 @@ namespace musica
...
+    /// @param dose_rates Output array for dose rates [photons/cm^2/s] (layer, reaction)
     /// @throws std::runtime_error if operation fails
-    void RunFromConfig(double *const photolysis_rate_constants, double *const heating_rates);
+    void RunFromConfig(double *const photolysis_rate_constants, double *const heating_rates, double *const dose_rates);
 
     /// @brief Get the number of photolysis reactions
     /// @return Number of photolysis reactions
     /// @throws std::runtime_error if operation fails
-    int GetPhotolysisRateCount();
+    int GetPhotolysisRateConstantCount();
```
boulderdaze, line 110: Just curious, why is `Count` changed to  `ConstCount`?

include/musica/tuvx/tuvx.hpp
```diff
@@ -31,12 +31,19 @@ namespace musica
     /// @param error Error struct to indicate success or failure
     void Create(const char *config_path, GridMap *grids, ProfileMap *profiles, RadiatorMap *radiators, Error *error);
 
+    /// @brief Create an instance of TUV-x from a JSON/YAML string
+    /// All parameters (solar zenith angle, Earth-Sun distance, atmospheric profiles, etc.)
+    /// are read from the JSON configuration file, similar to the Fortran tuvx.F90 driver
+    /// @param config_string JSON/YAML configuration string
+    /// @throws std::runtime_error if operation fails
+    void CreateFromConfigString(const char *config_string);
+
```
boulderdaze, line 40: What are JSON/YAML configuration strings? Does a string refer to a filename? Since both APIs have the same signature (const char*), this seems confusing to me.

boulderdaze (changes requested):

include/musica/tuvx/tuvx.hpp
mattldawson: The string contains the actual JSON/YAML code. Something like:
```
json_string = "{ "foo": 23, "bar": false }"
```

include/musica/tuvx/tuvx.hpp
boulderdaze: I understand it, thank you!

include/musica/tuvx/tuvx.hpp
boulderdaze: I got it, thank you!

PR #606, opened 2025-09-12
Add TUV-x Grid/GridMap Python classes

### LINKED ISSUE #597: Expose the TUV-x grid and grid map classes through the Python API

mattldawson:
This PR adds the ability to create and use TUV-x grids and grid maps in Python. It makes use of the now public data members of the TUV-x `grid_t` class to directly modify the collection of grids in a `grid_warehouse_t` so that errors can be properly raised and to fill in gaps of functionality of the `grid_warehouse_t` class (e.g., removing grids, finding grids by index).

An attempt was made to have the `GridMap` class mimic  the syntax of a Python dictionary in anticipation of porting TUV-x to C++/Python.

The error handling for grids and grid maps in the C/C++ library was updated to return more meaningful error messages.

closes #597


musica/grid.cpp
```diff
@@ -0,0 +1,206 @@
...
+            }
+            return result;
+          },
+          // Setter - converts numpy array to C++ array
+          [](musica::Grid &self, py::array_t<double, py::array::c_style | py::array::forcecast> array)
+          {
+            py::buffer_info buf = array.request();
+            musica::Error error;
+            size_t size = self.GetNumberOfSections(&error) + 1;
+            if (!musica::IsSuccess(error))
```
K20shores, line 127: This is yet another way to pass arrays from python to c++, and is different from what we did in CARMA, but I think I like this better

K20shores (approved):
I think I do like the monkey patch thing in the python classes. I'm hoping that we can replicate the same thing with the mechanism configuration objects

musica/grid.cpp
mattldawson: This was a Copilot suggestion, but I liked it too.

PR #632, opened 2025-10-03
Organize musica directory

boulderdaze:
in progress


K20shores (changes requested):
Would you mind splitting up `types.py` as well with this?

K20shores:
> > Would you mind splitting up `types.py` as well with this?
> 
> I was thinking that it might make more sense for it to be named micm instead of type, since including both the solver and state under type is a bit confusing. What do you think?

Yes, or maybe put it into a folder called micm and split it into micm, state, conditions? Either way is fine with me

boulderdaze:
> > > Would you mind splitting up `types.py` as well with this?
> > 
> > 
> > I was thinking that it might make more sense for it to be named micm instead of type, since including both the solver and state under type is a bit confusing. What do you think?
> 
> Yes, or maybe put it into a folder called micm and split it into micm, state, conditions? Either way is fine with me

Okay, I created an issue for this. I'll create an separate PR for this. 
-  https://github.com/NCAR/musica/issues/633

PR #769, opened 2026-02-02
703 package configurations with the musica python package

### LINKED ISSUE #703: Package configurations with the musica python package
Following #702 so it can be documented with the musica-cli, we need to figure out how to package up our configurations

**Acceptance criteria**
- Configurations are packaged with the musica python package
- Configurations are easily usable with the examples copied out with our CLI

**Ideas**
- Figure out how to include the [configs](https://github.com/NCAR/musica/tree/b9a9afac497f2e899832066f4babf19575c5029c/configs) with the python package and either
  - make the cli copy these configs to the same location as the script automatically,
  - or include a path to the config file like music-box does, [music-box examples](https://github.com/NCAR/music-box/tree/main/src/acom_music_box/examples) so that example configurations are easily accessible

K20shores:
Closes #703

- uses [`find_config_path`](https://github.com/NCAR/musica/blob/38ea742e55cd7e2e8b1dfa65761684bf50303622/python/musica/utils.py#L8) to find configurations throughout all of our tests, all of our examples. This means user won't have to copy the `configs` folder around since we package our configs with the wheel
- also adds a chapman box model with tuvx photo rates and corrects the configuration to have correct rates
- updates the fortran assert function to place the file number directly next to the filename to make it easy to navigate to errors from the command line

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -177,8 +177,8 @@ test-command = [
   "python ts1_box_model.py",
   "musica-cli -e TS1LatinHyperCube -o .",
   "python ts1_latin_hypercube.py",
-  # "musica-cli -e Chapman -o .", # for some reason, the netcdf file can't be saved in the linux wheel that tests this, but it works for ts1...
-  # "python chapman.py",
+  "musica-cli -e Chapman -o .",
+  "python chapman.py",
 ]
 
 # Set up pre-build hooks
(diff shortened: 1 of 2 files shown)


boulderdaze:
> > Since users don't necessarily need to know where the configuration root directory is, it may be difficult for them to modify the configuration file if they want to. Should we assume users won't modify the configuration?
> 
> I think they will want to modify the configuration. However, they still need them for the configurations to run, and we need to have our tests work either from wheels or editable installs (at least I feel like that's nice). So the trade off is that they would have to print the result of that find config thing and copy that. However, it means they can copy just the python file and run without worrying about finding the configuration file.
> 
> I'm open to suggestions though

I see. It will be a one-time inconvenience for users while we get the tests running. I am not sure which approach is better, but I'll approve this PR.

PR #852, opened 2026-04-21
Cmake changes for external collaborators

K20shores:
This PR contains changes needed to get musica to build for CATChem. All changes made here are in coordination with @zmoon


K20shores:
I let this get out of hand, splitting into one branch just to support noaa, another to address warnings as errors

PR #853, opened 2026-04-21
Update MUSICA installation method in binder to reflect changes in mechanism configuration reactant/product in tutorial

### LINKED ISSUE #849: Fix the bugs in notebook
The third notebook seems to be broken following the merge of PR #841. Investigate the issue.

boulderdaze:
When the API changes, the binder build can break because it points to `HEAD`.
This PR proposes pointing binder to the `release` branch so it remains stable and only updates with the official releases. 

Building from source seems noticeably slower than installing from pypi, so we'd like to use the pypi distribution whenever possible.

I confirmed the binder build works with the current branch. After it's merged into main, it gets merged into release, which the branch the binder clones. 
 
- Additional change: the minimum version of python should be consistent across the code base: `3.10` as defined in `pyproject.toml`
- Closes #849


K20shores (approved):
Fine with me. We used to install from pypi, but it is beneficial to be able to install a specific branch to test out features ahead of a release in binder. The biggest value add here is all of the setup to get an installation from source working in binder. If you want, you can add a line in `binder/postBuild` which runs `pip install musica[tutorial]` and just commenting out `pip install .[tutorial]` with a comment saying to uncomment that to test out binder on a particular branch

boulderdaze:
> Fine with me. We used to install from pypi, but it is beneficial to be able to install a specific branch to test out features ahead of a release in binder. The biggest value add here is all of the setup to get an installation from source working in binder. If you want, you can add a line in `binder/postBuild` which runs `pip install musica[tutorial]` and just commenting out `pip install .[tutorial]` with a comment saying to uncomment that to test out binder on a particular branch

Yeah I agree with you. If it's not about speed, building from the source make it easier to maintain. 
Currently we need to build from the local because the latest pypi doesn't include the API changes, but yes I'd like to have that for the next release.

K20shores:
@boulderdaze fine with me

boulderdaze:
> @boulderdaze fine with me

I added the comment for installation using pypi.

PR #856, opened 2026-04-24
codespell

zmoon:
* used it to find and correct some misspellings
* added a GHA workflow

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,24 @@
+name: codespell
+
+on: 
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  codespell:
+    name: Run codespell
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          submodules: recursive
+
+      - uses: codespell-project/actions-codespell@v2
+        with:
+          check_hidden: true
+          check_filenames: true
+          skip: "*.pdf,*.ipynb"
+          ignore_words_list: "inout,pres,aer"
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -21,4 +21,4 @@ jobs:
           check_hidden: true
           check_filenames: true
           skip: "*.pdf,*.ipynb"
-          ignore_words_list: "inout,pres,aer"
+          ignore_words_list: "inout,pres,aer,bui"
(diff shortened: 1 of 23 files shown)


K20shores (approved):
well this is neat, and a little embarrassing we had so many

zmoon:
oopsies few more to address/ignore

src/tuvx/interface_grid_map.F90
```diff
@@ -99,7 +99,7 @@ end subroutine internal_add_grid
 
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
-  function interal_get_grid(grid_map, c_grid_name, c_grid_name_length, &
+  function internal_get_grid(grid_map, c_grid_name, c_grid_name_length, &
```
zmoon, line 102: This should be the only change that is potentially consequential. But I didn't see it used anywhere.

K20shores: It is used, it's just bound to a c function call [`InternalGetGrid`](https://github.com/NCAR/musica/blob/a46c82af40755225106571f6df95e121390973cc/include/musica/tuvx/grid_map.hpp#L137). So you won't see a `call interal_get_grid` because it would only be called from C/C++ as `InternalGetGrid()`

zmoon: Oh gotcha. So this change is fine then.

K20shores: yes, it is fine

K20shores:
the test failure is expected because we can't run that test against forks

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,24 @@
...
+jobs:
+  codespell:
+    name: Run codespell
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          submodules: recursive
+
+      - uses: codespell-project/actions-codespell@v2
```
zmoon, line 19: One annoying thing about the Action is that it doesn't pin the version of codespell itself, so when new corrections are added, it'll start finding new things and failing the workflow unexpectedly. Usually I use it as a pre-commit hook, where you _can_ specify the version. But I could also just replace this part here with installing and calling it directly instead. The only benefit of the Action is that it provides annotations.

K20shores: I don't have a preference one way or the other. A certain amount of friction is good, and for something that can update our docs I think it might be appropriate to be a little annoyed. We can always update that in the future

zmoon: Either way. Just wanted to let you know what you were getting into!

boulderdaze: I see, thanks for the warning. I agree with Kyle. If a new version fixes something that was not an issue before, I think I'd want to see what changed and if it's worth it. If it ends up being a hassle, we can always install the specific version as you mentioned.

boulderdaze (approved):
Thank you!

boulderdaze:
I'll go ahead and merge it.

PR #859, opened 2026-04-28
Add TUV-x subset photolysis example and tutorial

K20shores:
Demonstrates how to reduce photolysis computation cost by filtering the TS1/TSMLT JSON config to only the reactions needed by a given mechanism, using the config_string API. Includes a %timeit comparison cell to quantify the runtime difference.


dwfncar (approved):
Looks good.

PR #863, opened 2026-04-28
installing with rpath

K20shores:
- add rpath for macOS, apparently necessary to work with uv
- add `constraint_init_iterations_` to fortran now that we are using a more updated micm version
- with the new micm verison, all rate constant classes were removed so we only need to pass paramteres

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,6 +1,5 @@
 FROM fedora:latest
 
-ARG MUSICA_GIT_TAG=main
 ARG BUILD_TYPE=Debug
 
 RUN dnf -y update \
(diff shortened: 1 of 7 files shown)


dwfncar (approved):

docker/Dockerfile
```diff
@@ -28,9 +27,8 @@ RUN cd musica \
              -B build \
              -D CMAKE_BUILD_TYPE=${BUILD_TYPE} \
              -D MUSICA_BUILD_FORTRAN_INTERFACE=ON \
-             -D MUSICA_GIT_TAG=${MUSICA_GIT_TAG} \
     && cd build \
-    && make install -j 8
+    && make install -j 1
```
boulderdaze, line 31: `make install` without the option should default to single-threaded execution.
```suggestion
    && make install
```

boulderdaze (approved):
Looks good. I just have a minor suggestion.

PR #878, opened 2026-05-14
Support for CATChem

### LINKED ISSUE #871: Support NOAA
Make sure we can compile with intel, c++20 from gcc 11

Also make sure that compile warnings as errors is added only in our CI tests, and also that it is off by default

K20shores:
- Bump micm to latest tag that only uses c++20
- Bump mechanism_configuration to latest tag that has fmt support
- Enable MECH_CONFIG_USE_FMT for NOAA model support
- turns on address sanitzer for docker builds
  - this caught a bug that was causing fortran to fail on windows. For some reason that showed up only with these changes. Asan showed that we were leaking memory sometimes and fixing those fixed the fortran test failures

Closes #871


docker/Dockerfile
```diff
@@ -1,9 +1,8 @@
-FROM fedora:latest
+FROM fedora:42
```
boulderdaze: Pinning to fedora42 is okay but it's inconsistent with the other Dockerfiles. Do we want to pin all of the images to the specific tag?

docker/Dockerfile
```diff
@@ -13,8 +12,8 @@ RUN dnf -y update \
         git \
         hdf5-devel \
         lapack-devel \
-        libcurl-devel \
-        m4 \
+        libasan \
+        lcov \
```
boulderdaze: I don't think lcov needs to be installed, since this docker has nothing to do with coverage testing.
I also think we can remove valgrind.

As for removing `libcurl` and `m4`, `libcurl` is for linking against netcdf, and `m4` is a build tool that would be potentially invoked when musica fetches and compiles projects from source. Since the docker builds are succeeding, it looks like we could drop them.

docker/Dockerfile
```diff
@@ -30,16 +29,24 @@ COPY . musica
...
-             -D MUSICA_ENABLE_TESTS=ON \
+        -DCMAKE_C_FLAGS="-fsanitize=address -g" \
+        -DCMAKE_CXX_FLAGS="-fsanitize=address -g" \
+        -DCMAKE_Fortran_FLAGS="-fsanitize=address -g" \
+        -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address" \
+        -DCMAKE_SHARED_LINKER_FLAGS="-fsanitize=address" \
+        -DMUSICA_BUILD_FORTRAN_INTERFACE=ON \
+        -D MUSICA_ENABLE_TESTS=ON \
+        -B build \
+        -D CMAKE_BUILD_TYPE=${BUILD_TYPE} \
```
boulderdaze: Both `-DCMAKE`, `-D CMAKE` are used in the same command. It would be nice to have them consistent.
 

```suggestion
        -DMUSICA_BUILD_FORTRAN_INTERFACE=ON \
        -D MUSICA_ENABLE_TESTS=ON \
        -D CMAKE_BUILD_TYPE=${BUILD_TYPE} \
        -D CMAKE_C_FLAGS="-fsanitize=address -g" \
        -D CMAKE_CXX_FLAGS="-fsanitize=address -g" \
        -D CMAKE_Fortran_FLAGS="-fsanitize=address -g" \
        -D CMAKE_EXE_LINKER_FLAGS="-fsanitize=address" \
        -D CMAKE_SHARED_LINKER_FLAGS="-fsanitize=address" \
        -B build \
```

boulderdaze (changes requested):

boulderdaze (approved):

PR #882, opened 2026-05-19
Address warnings as errors

### LINKED ISSUE #879: Setup clang tidy
During #852, I started addressing all compiler warnings as errors. All of that work still lives on the branch https://github.com/NCAR/musica/tree/warnings_as_errors. It wasn't actually necessary to do that work to support noaa, but it came out as a consequence as I was trying to [remove the compile flag for `-Werror`](https://github.com/NCAR/musica/pull/878/changes/db631563562c5628c6a51f526421130456d7f82e)

Then in addressing the warnings as errors, I was reminded in micm that we have clang tidy. We will use that instead

**Acceptance criteria**
- clang tidy actions now cause CI failures
- warnings as errors are turned on
- all warnings failures are addressed

**Ideas**
-

K20shores:
closes #879 

- depends on https://github.com/NCAR/micm/pull/984
- Adds `MUSICA_COMPILE_WARNING_AS_ERROR` and sets it to on for all of our CI tests
- Addresses all warnings so that they no longer show up across any of our compilers
- adds an intel dockerfile build
- changes how we allocate/deallocate version information. ifx did not work with what we were doing


K20shores:
going to retask #879 to use clang tidy instead of this. This will allow us to use the same clang tidy file for all of our repositories and will be good enough

PR #884, opened 2026-06-01
Refactor build system, cleanup docker configuration and fortran components

### LINKED ISSUE #869: Rewrite the CMake configuration to avoid FetchContent for MUSICA
In #864, the fetch-content github action was temporarily disabled due to a linking issue. The root cause is that a library built with one toolchain/ABI is being linked with another, leading to unresolved symbols.

Specifically, MUSICA was compiled on Fedora using GCC (libstdc++), but later linked in an Ubuntu environment using IntelLLVM (icpx). Because these use different C++ standard libraries/ABIs, required `<format>` symbols from the GCC-built library aren’t resolved during linking.

Additionally, for local testing of musica-fortran, the current CMake setup fetches musica from `main` branch, which is not in sync with local source changes, making it difficult to run tests locally.

This issue proposes:
- Rewriting the CMake configuration to avoid fetch-content for MUSICA
- Re-enabling the currently disabled github action workflow.

boulderdaze:
This PR:
- Refactores the CMake build system
  - Removes the use of `FetchContent` for `musica-fortran`
  - Removes a sub `project()` for `musica-fortran`
    - `musica-fortran` is not a standalone cmake project because it depends on the variables defined by the parent project (`MUSICA_MOD_DIR`, `MUSICA_LIB_DIR`, etc.) and sub project only adds a library via `add_library(musica-fortran)`.
    - The project versioning is not useful since the project is included via `add_subdirectory()`
   
  - Changes the default value of `MUSICA_ENABLE_TESTS` from `ON` to `OFF`. 
    - This makes `musica` easier to consume as a dependency in the downstream without the need to explicitly disable the test builds.

- Fixes and cleans up Docker configuration
  - Fixes build type for coverage (Release -> Debug) 
  - `Docker.openmp` relies on MPI, which we already have its own dedicated dockerfile (`Dockerfile.mpi_openmp`). Removes MPI pieces and makes it OpenMP only. 
  - `Dockerfile.wheel` has a wheel filename hardcoded to a specific version. This requires manually updating the line at every release. Set it as an argument.
  - Merges `Dockerfile.fortran-gcc.integration` into the main`Dockerfile` where it already builds `musica-fortran` and run its test suite using gcc.
    - Also removes the smoke test (`/musica/src/test/test_simple.c`) which is redundant with `MUSICA_ENABLE_TEST=ON`.  The test suite provides broader coverage. 
  - Updated Intel and NVHPC compiler builds
    - `musica` and `musica-fortran` are now built using their respective native compilers
    - This follows the removal of JIT compiler support.
  
- Clean up
  - Removes the fortran demo program, as the existing test suite can serve better reference. 

- Closes #869


CMakeLists.txt
```diff
@@ -130,6 +130,7 @@ endif()
 ################################################################################
 # MUSICA-Fortran
 if(MUSICA_BUILD_FORTRAN_INTERFACE)
+  enable_language(Fortran)
```
K20shores: Maybe we merge this with the if statement [above](https://github.com/NCAR/musica/blob/2b1bbbf35bf6617cf3fc94069f6ef476d250277d/CMakeLists.txt#L68-L73)

cmake/CodeCoverage.cmake
```diff
@@ -158,7 +158,9 @@ elseif(NOT CMAKE_COMPILER_IS_GNUCXX)
     endif()
 endif()
 
-set(COVERAGE_COMPILER_FLAGS "-g -O0 -fprofile-arcs -ftest-coverage -fcheck=bounds,do,pointer -ffpe-trap=zero,overflow,invalid"
+set(COVERAGE_COMPILER_FLAGS "-g -O0 -fprofile-arcs -ftest-coverage"
+    CACHE INTERNAL "")
+set(COVERAGE_Fortran_FLAGS "-fcheck=bounds,do,pointer -ffpe-trap=zero,overflow,invalid"
```
K20shores, line 163: fortran only flags?

docker/Dockerfile.docs
```diff
@@ -1,18 +1,16 @@
...
-ARG BUILD_TYPE=Release
-
 RUN dnf -y update \
     && dnf -y install \
         cmake \
         doxygen \
         gcc \
         gcc-c++ \
-        gfortran \
+        gcc-gfortran \
```
K20shores, line 13: is there a difference between this and gfortran?

docker/Dockerfile.mpi
K20shores, line 1: Regarding OpenMP and MPI, we don't actually have any tests that really use this. The only component of music that uses MPI right now is Tuvox, and we don't test it here. Just a comment

docker/Dockerfile
```diff
@@ -1,38 +1,40 @@
...
              -D MUSICA_BUILD_FORTRAN_INTERFACE=ON \
+             -D MUSICA_ENABLE_TESTS=ON \
     && cd build \
-    && make install -j 1
-    
-# Try building a simple application
-RUN cd musica/build \
-    && gcc $(pkg-config --cflags musica) -o test_simple /musica/src/test/test_simple.c $(pkg-config --libs musica) \
-    && ./test_simple
+    && make install -j$(nproc)
```
K20shores, line 38: This is actually useful. What it tests is that our configuration can be properly found by pkg-config. I think it would actually be nice to include this in every dockerfile that does an install with a different compiler

docker/Dockerfile.intel
K20shores, line 1: since this is now testing all of musica, let's rename to docker/Dockerfile.intel

docker/Dockerfile.fortran-nvhpc.pkgconfig
K20shores, line 1: since this is now testing all of musica, let's rename to docker/Dockerfile.nvhpc

K20shores (changes requested):
Couple of changes I think make sense given the changes here.

If we are removing the fortran `project`, it would also make sense to remove every other `project`
- [javascript](https://github.com/NCAR/musica/blob/2b1bbbf35bf6617cf3fc94069f6ef476d250277d/javascript/CMakeLists.txt#L4)
- [julia](https://github.com/NCAR/musica/blob/2b1bbbf35bf6617cf3fc94069f6ef476d250277d/julia/CMakeLists.txt#L1)

And, if we are removing the projects, then we can probably [remove the src project](https://github.com/NCAR/musica/blob/2b1bbbf35bf6617cf3fc94069f6ef476d250277d/src/CMakeLists.txt#L4-L8) and change the name of the [top level project](https://github.com/NCAR/musica/blob/2b1bbbf35bf6617cf3fc94069f6ef476d250277d/CMakeLists.txt#L4) from `musica-distribution` to `musica`

CMakeLists.txt
boulderdaze: Yes, done

docker/Dockerfile
boulderdaze: Yes, I added this smoke test back to those dockerfiles: `Dockerfile`,  `Dockerfile.fortran-intel.pkgconfg`, and `Dockerfile.fortran-nvhpc.pkgconfig`

docker/Dockerfile.docs
boulderdaze: Yeah, unlike ubuntu where the fortran package name is the same as the binary name (`gfortran`), on fedora `gcc-gfortran` is the package name.

cmake/CodeCoverage.cmake
boulderdaze: I learned that `-fcheck=bounds,do,pointer` and `-ffpe-trap=zero,overflow,invalid` are gfortran-specific flags and not valid for C/C++ compilers. Since the flag `COVERAGE_COMPILER_FLAGS` gets applied to all three languages, fortran specific flag is created for them.

docker/Dockerfile.fortran-nvhpc.pkgconfig
boulderdaze: I also considered renaming the dockefiles but decided against it for a few reasons. 
There is already a `Dockerfile.nvhpc` which builds the musica C/C++ with GPU support (`MUSICA_GPU_TYPE=a100`). It disables TUV-x and CARMA and does not build the fortran interface. 

The intel and nvhpc pkg-config builds both compile musica and fortran interfaces with their respective compilers and it also validates pkg-config setup. These builds do not include GPU support, but they do enable fortran interface and tests, and they run a pkg-config smoke test for `musica-fortran`. 

Given these differences, I think the current filenames still reflect their purposes and can remain as they are. What do you think?

docker/Dockerfile.intel
boulderdaze: See the comments below

docker/Dockerfile.mpi
boulderdaze: Ah I see, that makes sense

docker/Dockerfile.fortran-nvhpc.pkgconfig
K20shores: Intel can't have a gpu option, and we only use nvhpc specically because we want the gpu option. Mostly I'm a fan of consolidating our docker builds for nvhpc and intel because realistically we will always build the c++/fortran APIs for these compilers (since they target our HPC use cases which will ilkely always need fortran), and the GPU option should always be on for nvhpc regardless of which interface is built.

I also want to do better about making our test run relatively fast. Anything we can do to reduce CI build times seems smart.

That said, I'm happy to leave them as they are since the bottleneck is really the python wheel building

boulderdaze:
The following changes were made in response to the code review comments:

- Removed sub-projects: `project(musica-wasm)`, `project(musica-julia)` `project(musica)` in `src/CMakeLists.txt`
- Replaced `musica-distribution` with `musica`
- Updated `${CMAKE_BINARY_DIR}/lib` to `${MUSICA_LIB_DIR}` for build output directories
- Updated hardcoded lib to `${CMAKE_INSTALL_LIBDIR}` for install destinations

- `fortran/packaging/CMakeLists.txt` now uses the dedicated `musica-fortranConfig.cmake.in` template, which uses `musica-fortran` for the exports file and `check_required_components`, and calls `find_dependency(musica REQUIRED)` to pull in the C/C++ library automatically

K20shores (approved):

docker/Dockerfile.fortran-nvhpc.pkgconfig
boulderdaze: Ah, I am fine with consolidating the docker builds. Do you know how we should handle the TUV-x and CARMA build options with GPU support? The nvhpc image specifically only tests the GPU MICM path.

K20shores: I think we should be able to turn them on. I'd say try to turn them on and if they fail, turn them off with a comment about what failed

boulderdaze: Okay I will give it a try. If it fails, would our plan be to keep them separate as they are now instead of consolidating them?

PR #903, opened 2026-06-21
Add config string constructor, species properties to julia interface

### LINKED ISSUE #902: Updates to julia interface
In developing clima with musica, @obin1 found that we have a couple of holes in our API that we need to fill to make the julia interface more useable

**Acceptance criteria**
- Some equivalent of get_species_property is defined
- a constructor that takes a json/yaml string should be added

**Ideas**
- copy the fortran interface

K20shores:
Closes #902 

- allows species properties to be returned, which will be needed when converting from mixing ratio to molar concentration
- allows a mechanism to be built from a string config
- also allows getting both the micm and musica versions, changes the name of `get_version` to `get_musica_version`
- adds tests for all of these

### DIFF
--- a/julia/bindings/musica_julia.cpp
+++ b/julia/bindings/musica_julia.cpp
@@ -170,6 +170,24 @@ JLCXX_MODULE define_julia_module(jlcxx::Module& mod)
       });
 
   mod.method(
+      "cpp_create_solver_from_string",
+      [](const std::string& config_string, int solver_type)
+      {
+        musica::Error error;
+        musica::MICM* micm =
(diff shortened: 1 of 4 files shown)


sjsprecious (approved):
LGTM. Shall we let @obin1 test it in CliMA and tell us whether this works, or it is not ready yet?

boulderdaze (approved):

K20shores:
> LGTM. Shall we let @obin1 test it in CliMA and tell us whether this works, or it is not ready yet?

Hmmmm I don't know how easy that would be... I think if @obin1 does a local installation he can make julia pick up what he built, but it may be better to wait for the release for next week

PR #919, opened 2026-06-30
Use refactored mechanism configuration

### LINKED ISSUE #917: update to latest mech config tag
Now that mechanism configuration has been refactored, we need to bring it in to prepare to use the new aerosol and emissions work in coming

**Acceptance criteria**
- latest mech config tag is used
- aerosol work disabled (for now) until we bring in the aerosol stuff

K20shores:
Closes #917 

- disables MIAM for now in python, since we are waiting on the new config information
- uses the refactored mechanism configuration
  - since we parse to a single mechanism, we no longer need the v1/v0 idea
    - I'm going to rename v1_parse.cpp after this PR gets in so that it's easy to see what changed
- updated the python wrapper to use the new mech config


boulderdaze (approved):
Looks cleaner!

Issue #21, opened 2023-05-16
Add option to link to MICM library from CAMDEN


mattldawson:
- [x] NCAR/musica#22
- [x] #23
- [x] #24
- [x] Testing on Cheyenne

PR #929, opened 2026-07-07
Add MIEM emissions integration (parse, translate, build wiring)

smeechncar:
## Summary

Establishes the translation pipeline from MechanismConfiguration's emissions schema, through musica, into MIEM's runtime types, plus baseline end-to-end tests against real data.

- **Translation layer**: `musica::ReadEmissionsConfiguration` / `ReadEmissionsConfigurationFromString` parse a MechanismConfiguration emissions section, and `ConvertEmissions` translates the parsed types into MIEM's runtime types (`miem::Source` + `miem::Regridding`) — covering source mode, source type, temporal interpolation, vertical injection, regridding type, and species-map scaling.
- **Build wiring**: MIEM is added as an opt-in dependency (`MUSICA_ENABLE_MIEM`, default OFF), pinned to miem `main` (includes the upstream CMake module-path fix and the NOx fixture, NCAR/miem#20 and #21).
- **Baseline real-data tests**: two end-to-end tests that don't mock anything — each parses a real config, translates it, and hands the result to MIEM's actual `EmissionsBuilder::Build()`/`Run()` against a real committed NetCDF fixture:
  - BC: CAMS-GLOB-ANT black carbon subset
  - NOx: 2024 anthropogenic NOx subset, exercising a one-to-many species split (`nox_anth_sum` → `NO` + `NO2`), asserted against the exact configured scaling ratio
  - Plus a 10-case parser unit test suite covering error paths (missing emissions section, unresolved inventory/species-map references, bad file paths) and enum-mapping correctness.

## Direction

These two tests establish a repeatable pattern — real config → translate → real MIEM build/run against a real fixture — rather than a one-off. The intent is to grow coverage the same way: drop a new subset fixture into `miem/test/data/`, write a small config, and add one more `end_to_end_*.cpp` following `end_to_end_nox.cpp`. Reviewers should read this PR as the first two data points in that series, not the full extent of intended fixture coverage.

## Test plan
- [x] `ctest --test-dir build -R miem` — 10/10 `miem_parser` sub-tests, `miem_end_to_end` (BC), `miem_end_to_end_nox` (NOx) all pass
- [x] Verified against a clean FetchContent build with no local-source override, confirming the miem re-pin resolves cleanly from `origin/main`


smeechncar:
I hope this isn't too much this round. I really wanted to have some sort of end to end testing with real file subsets to verify that things are moving along correctly. I want to continue to add more of these types of tests with different real subset input files from UPTEMPO which will live in MIEM, similar to how this chain is working.

K20shores (changes requested):
Before I review, please rebase onto the branch in this PR (https://github.com/NCAR/musica/pull/926). It may require a bit of a refactor, but I think we are in general consensus that the work in #926 is the right direction

smeechncar:
@K20shores OK I have rebased on to that PR. The new MIEM additions still live outside of the new more generalized way of handling MICM. Next should I get the MIEM references to actually build on top of the new musica::configuration instead of how it is currently built on the outside of this? Is this the refactor you are referring to?

cmake/dependencies.cmake
```diff
@@ -67,7 +67,7 @@ endif()
 
 if(MUSICA_BUILD_C_CXX_INTERFACE AND NOT MUSICA_USE_PREBUILT)
   set_git_default(MECH_CONFIG_GIT_REPOSITORY https://github.com/NCAR/MechanismConfiguration.git)
-  set_git_default(MECH_CONFIG_GIT_TAG v2.0.0)
+  set_git_default(MECH_CONFIG_GIT_TAG a8c7c0de74800956ea6740261c8af240c1accb5b)  # main, includes emissions config (#281); swap to a tag once one is cut
```
K20shores, line 70:
```suggestion
  set_git_default(MECH_CONFIG_GIT_TAG a8c7c0de74800956ea6740261c8af240c1accb5b)
```

cmake/dependencies.cmake
```diff
@@ -135,6 +135,26 @@ if (MUSICA_ENABLE_MICM AND MUSICA_BUILD_C_CXX_INTERFACE AND NOT MUSICA_USE_PREBU
   FetchContent_MakeAvailable(micm)
 endif()
 
+################################################################################
+# MIEM
+# Skip if using prebuilt musica (already includes miem)
+
+if (MUSICA_ENABLE_MIEM AND MUSICA_BUILD_C_CXX_INTERFACE AND NOT MUSICA_USE_PREBUILT)
+  set_git_default(MIEM_GIT_REPOSITORY https://github.com/NCAR/miem.git)
+  set_git_default(MIEM_GIT_TAG 3cc9ff9684dfdc13a49ab327ce2539bc223b1fd4)  # main; no tags exist yet
```
K20shores, line 144:
```suggestion
  set_git_default(MIEM_GIT_TAG 3cc9ff9684dfdc13a49ab327ce2539bc223b1fd4)
```

include/musica/configuration/emissions.hpp
K20shores, line 1: I think the reading could/should be merged with the rest of the configuration reading. What do you think @smeechncar, @boulderdaze?

src/configuration/emissions.cpp
K20shores, line 1: Oh, I don't know. It also makes sense to have this here. Maybe what we should actually do is have `musica/configuration` read the config file, and then pass off to micm/miem?

K20shores (commented):
Nothing wrong necessarily, just wondering if we should do some reorganization

src/configuration/emissions.cpp
smeechncar: I'm OK with the reorganization. We unfortunately both checked these branches out last Friday and worked on them over the weekend. If I had waited until this week, I would have blended it in with your branch anyway to match its current style. Just let me know which direction you want to go in. If needed, I could wait until #926 is merged into main since it is looks like it is about to be approved.

boulderdaze:
Sorry, I haven't received any email notification for this PR. I didn't know it is waiting for my review. 

Just seeing Kyle ask you to rebase this PR onto his open PR makes me realize that I didn't receive the request for re-review. if my delayed review made both of you to wait on me, sorry about that.

@K20shores, I'll go ahead and merge your PR so it's easier to review the changes this PR introduces.
@smeechncar I merged `main` into your branch so I could see only the file diffs that your branch introduces. I hope you don't mind!

CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ option(MUSICA_ENABLE_MEMCHECK "Enable memory checking" OFF)
 option(MUSICA_BUILD_DOCS "Build the documentation" OFF)
 option(MUSICA_ENABLE_MICM "Enable MICM" ON)
 option(MUSICA_ENABLE_MIAM "Enable MIAM" OFF)
+option(MUSICA_ENABLE_MIEM "Enable MIEM emissions runtime" OFF)
```
boulderdaze: The runtime description feels a bit out of context. It doesn't have to be phrased the same way as the other options, but if you are adding extra information, it's worth considering whether the information should be included for the other options as well. 
If `enable micm` on, it's also built and available at runtime. Adding this description to only one option feels a bit inconsistent and could be confusing.

src/test/unit/miem/end_to_end.cpp
```diff
@@ -0,0 +1,92 @@
...
+    return yaml.str();
+  }
+}  // namespace
+
+TEST(MiemEndToEnd, MechanismConfigThroughMusicaToRealMiemFlux)
+{
+  musica::Emissions parsed = musica::ReadEmissionsConfigurationFromString(EmissionsConfigYaml());
+  ASSERT_EQ(parsed.sources.size(), 1);
+
+  miem::Emissions module =
```
boulderdaze: It is a small detail, but would it make sense to name the variable `miem::Emissions emissions`? 
`emissions` feels a bit more descriptive while `module` seems a little too generic.
```suggestion
  miem::Emissions emissions =
```

include/musica/configuration/emissions.hpp
boulderdaze: Is the emissions configuration meant to be independent from reactions, aerosols, and species? Looking at the example configuration, it seems like species and phases can also be empty. 

Since aerosols will be incorporated into the current parser as well because gas phase and aerosol particles can be simulated together, I think this question may help us consider whether having the emissions parsers in the same place makes sense as well.

smeechncar:
> Sorry, I haven't received any email notification for this PR. I didn't know it is waiting for my review.
> 
> Just seeing Kyle ask you to rebase this PR onto his open PR makes me realize that I didn't receive the request for re-review. if my delayed review made both of you to wait on me, sorry about that.

No apology necessary @boulderdaze!  

> @K20shores, I'll go ahead and merge your PR so it's easier to review the changes this PR introduces. @smeechncar I merged `main` into your branch so I could see only the file diffs that your branch introduces. I hope you don't mind!

Merging back was on my list anyway, so thanks for doing that!

include/musica/configuration/emissions.hpp
K20shores: An idea behind mechanism configuration is that it's a configuration that describes chemistry, not a particular package. It can be used to configure miem/miam/micm, however it could be used to configure something else as well. It's an artificial barrier, but what it means is that we would be able to process a configuration file and describe the chemistry without needing to configure any of the underlying tools and a chemist might be able to understand the science without needing to know the implementation.

Because of that, I strongly suggest we merge the parsing here with the configuration. Then, we will need to pass down the structs to each package as appropriate for them to be configured.

src/configuration/emissions.cpp
K20shores: Would you make a pass at the reorganization? I imagine it as confiugre (creates a bunch of structs), then we dispatch the structs to each package

CMakeLists.txt
smeechncar: dropped "runtime" so it reads consistently with the MICM/MIAM option descriptions above it.

src/test/unit/miem/end_to_end.cpp
smeechncar: renamed to emissions in both this test and the NOx variant (end_to_end_nox.cpp) for consistency.

src/configuration/emissions.cpp
smeechncar: OK here is the first pass. Some things are just moving code around. Probably the biggest architectural change is the src/CMakeLists.txt in which before configuration/ was like a MICM helper and only existed when MICM was turned on. Now the directory always exists and contains module specific things as well as the read_mechanism, which is a reader for any package. Any new module could add its own Read/Convert pair here and reuse read_mechanism. 

See highlighted specifics below:

- `include/musica/configuration/read_mechanism.hpp` + `src/configuration/read_mechanism.cpp`: new shared `ReadMechanism`/`ReadMechanismFromString`, the single parse-or-throw step (no micm/miem dependency), returning the raw `mechanism_configuration::Mechanism`.
- `src/configuration/parse.cpp`: `ReadConfiguration`/`ReadConfigurationFromString` are now thin wrappers — `ConvertMechanism(ReadMechanism(path))`. `ConvertMechanism` and all its converters are unchanged, so every existing caller (C API, Python, JS/WASM, the `MICM` C++ ctors) keeps working with zero signature changes.
- `include/musica/configuration/emissions.hpp` + `src/configuration/emissions.cpp`: moved over from `include/musica/miem/` + `src/miem/` — same `Emissions` struct, same `ReadEmissionsConfiguration`/`ReadEmissionsConfigurationFromString`/`ConvertEmissions` signatures, now also thin wrappers over `ReadMechanism`/`ReadMechanismFromString`.
- Deleted `src/miem/` and `include/musica/miem/` entirely — there were no external callers besides our 3 unit test files (no C API/Python/Fortran/JS binding for miem exists yet), so this was a clean move.
- Had to also move `add_subdirectory(configuration)` in `src/CMakeLists.txt` out of the `MUSICA_ENABLE_MICM` guard (it was previously nested inside it) so `configuration/` — now shared infrastructure — still builds when MIEM is on and MICM is off. `configuration/CMakeLists.txt` gates `parse.cpp` behind `MUSICA_ENABLE_MICM` and `emissions.cpp` behind `MUSICA_ENABLE_MIEM`, with `read_mechanism.cpp` always built.

smeechncar: However, I realize now that #930 touches a lot of this differently. That could complicate this.

include/musica/configuration/emissions.hpp
```diff
@@ -0,0 +1,23 @@
...
+{
+  struct Emissions
+  {
+    std::vector<miem::Source> sources;
+    miem::Regridding regridding;
+  };
+
+  Emissions ReadEmissionsConfiguration(const std::string& config_path);
+  Emissions ReadEmissionsConfigurationFromString(const std::string& json_or_yaml_string);
+  Emissions ConvertEmissions(const mechanism_configuration::Mechanism& mechanism);
```
K20shores, line 22: I still think we'd be able to have a single `Read*` function. Rather than returning `Emissions`, it would return a mechanism and then somewhere else in musica we'd dispatch to micm/miam/miem, as appopriate

include/musica/configuration/read_mechanism.hpp
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include <mechanism_configuration/mechanism.hpp>
+
+#include <string>
+
+namespace musica
+{
+  mechanism_configuration::Mechanism ReadMechanism(const std::string& config_path);
+  mechanism_configuration::Mechanism ReadMechanismFromString(const std::string& json_or_yaml_string);
```
K20shores, line 10: Yeah, this is what I was expecting, which makes me wonder why we need `configuration/emissions.hpp`

src/configuration/emissions.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+  Emissions ReadEmissionsConfiguration(const std::string& config_path)
+  {
+    return ConvertEmissions(ReadMechanism(config_path));
+  }
+
+  Emissions ReadEmissionsConfigurationFromString(const std::string& json_or_yaml_string)
+  {
+    return ConvertEmissions(ReadMechanismFromString(json_or_yaml_string));
+  }
```
K20shores, line 156: I think it's mostly that I don't want emissions to read a configuration like this, because then we wouldn't ever use the rest of the mechanism. In my mind it makes sense to have one pathway for parsing, but perhaps I'm wrong

K20shores (commented):
I think this is in the right direction, but there's still some duplication that exists. What do you think?

src/configuration/emissions.cpp
boulderdaze: I think what Kyle meant was that `Chemistry` should include emissions? but I don't know, maybe it requires more work than just introducing emissions, and it could be beyond the scope of this PR. 
My vote is to accept this PR as it is, and to create a separate issue to plumb emissions to chemistry.

```C++
  struct Chemistry
  {
    micm::System system;
    std::vector<micm::Process> processes;
    // something like this
    Emissions emissions;  
  };
   
  // read configuration function returns chemistry
  Chemistry ReadConfiguration(const std::string& config_path);
  Chemistry ReadConfigurationFromString(const std::string& json_or_yaml_string);
  Chemistry ConvertMechanism(const mechanism_configuration::Mechanism& mechanism);
```

smeechncar: I also think that the plumbing questions could be handled in a follow on PR. I'm not 100% sure how to leave this current work though. Since the latest commit was in the right direction, maybe would _could_ just leave it as is (mostly) and refine this in a follow on PR?

K20shores: Not necessarily, I didn't have a better name at the time than `Chemistry`. I meant that we read a mechanism and then dispatch that into `Chemistry`, `Emissions`, `Aerosols`

boulderdaze: ah I see

boulderdaze (approved):
I'll approve this PR, and let's track what we discussed in a follow-up issue.

K20shores:
> I'll approve this PR, and let's track what we discussed in a follow-up issue.

I'd kinda rather get it done in this PR. But if the very next thing we work on is this issue, we can defer to that.

smeechncar:
I can take a stab at simplifying this via having a single Read* function and adjusting everything down stream from that based on your latest comments on read_mechanism and emissions.cpp/.hpp.

src/configuration/emissions.cpp
boulderdaze: You want to read a mechanism once and then dispatch that into Chemistry, Emissions, Aerosols. Chemistry itself has its own read configuration function, which then having this specific emissions parser makes sense. 
The refactoring you're requesting seems beyond the scope of this PR to me. 

```C++
  Chemistry ReadConfiguration(const std::string& config_path);
  Chemistry ReadConfigurationFromString(const std::string& json_or_yaml_string);
  Chemistry ConvertMechanism(const mechanism_configuration::Mechanism& mechanism);
```

smeechncar: I'm happy to do whatever you both would like. Just want to make sure we are all on the same page.

K20shores:
> I can take a stab at simplifying this via having a single Read* function and adjusting everything down stream from that based on your latest comments on read_mechanism and emissions.cpp/.hpp.

Oh, I'm convinced. I'll write up an issue explaining what I'm thinking and we can iterate on that.

K20shores (approved):

Issue #896, opened 2026-06-15
Evaluate use of `miam_config` namespace for consistency


boulderdaze:
The file `include/musica/miam/miam_types.hpp` defines the `miam_config` namespace but MUSICA does not currently use dedicated namespaces for its other subprojects.

- Evaluate whether retaining the `miam_config` namespace is necessary. Removing it may improve consistency with the rest of the codebase

- Once the mechanism configuration for aqueous reactions is in place, these structs will be removed and replaced by the corresponding mechanism configuration definitions.
```C++
    /// @brief Henry's law constant: HLC(T) = HLC_ref * exp(C * (1/T - 1/298.15))
    struct HenryLawConstant
    {
      double hlc_ref;   ///< Reference HLC at 298.15 K [mol m-3 Pa-1]
      double c = 0.0;   ///< Temperature dependence factor [K]
    };
```

- `struct SpeciesDef` seems superfluous given species, phase species. Check every type and see how many of them could be removed entirely

Issue #890, opened 2026-06-12
Make MIAM use species objects instead of string names


boulderdaze:
MIAM references species by name, while mechanism_configuration takes object references. MIAM treats the name as a lookup key into a shared namespace that gets resolved later.

```C++
// mechanism_configuration takes object references
so2 = mc.Species(name="SO2")
gas = mc.Phase(name="gas", species=[so2, h2o2, o3])   // <- objects

// MIAM  takes string names
DissolvedEquilibriumConstraint(reactant_names=["H2O"], product_names=["Hp", "OHm"], ...)
DissolvedReaction(reactant_names=["SO2OOHm", "Hp"], ...)  // <- strings
```

Accept species objects and resolve them at construction.
- This change makes the MUSICA APIs more consistent.
- Using objects instead of strings allows validation at construction time and helps catch errors such as typos early.

```C++
DissolvedEquilibriumConstraint(
    reactants=[hp, ohm],     // <- pass the mc.Species objects directly
    algebraic_species=ohm,
    ...
)
```

PR #930, opened 2026-07-08
Reorganize Python mechanism configuration types after adding aerosols

boulderdaze:
This PR:
- Introduces python bindings for the aerosol types.
- Reorganizes the binding functions and files into species, reactions, aerosol, as mirroring the mechanism configuraiton structure. This improves maintainability, as we continue adding aerosols and emissions. 
  - `common.cpp` calls `bind_mechanism_configuration`, which now internally calls `bind_species`, `bind_reactions` and `bind_aerosol`. 
- Removes the reassignment of `Version` which is currently  both imported and reassigned to the C++ type, eliminating the reassignment warning.
- Renames `_cpp_types`  with `_mc` since the `_cpp_types` is slight misleading. The module is bound to  `mechanism_configuration` rather than c++ types, and `_mc` serves as a short handle for the module.
- Is part of the ongoing work for issue #890 and #896


boulderdaze:
Hi @smeechncar, I am tagging you here just for visibility, since you're also working on bringing emissions types into musica. No need to review this PR, but I thought it it might be helpful for you to see what's coming.

K20shores (changes requested):
Looks good, but I think you need to enable miam to test these out. Look at [linux](https://github.com/NCAR/musica/blob/6a88a9ad03b1f2e6c0c921587ee606cbaffa6139/python/tools/prepare_build_environment_linux.sh#L61), [macos](https://github.com/NCAR/musica/blob/6a88a9ad03b1f2e6c0c921587ee606cbaffa6139/python/tools/prepare_build_environment_macos.sh#L47), also [pyproject.toml](https://github.com/NCAR/musica/blob/6a88a9ad03b1f2e6c0c921587ee606cbaffa6139/pyproject.toml#L72)

K20shores (changes requested):
One more thing, please enable the notebooks to run now that miam can be used again. Look at [this file](https://github.com/NCAR/musica/blob/25df47c4237c46b661064ec6a1c91db271690550/.github/workflows/notebooks.yml#L46-L48)

K20shores (approved):

PR #940, opened 2026-07-14
Consolidate configuration parsing onto a single read path

### LINKED ISSUE #932: Refactor configuration parsing
In #929, miem configuration parsing was brought in, but we have room to reduce redundancy.

We can read configuration files in two places

1. https://github.com/NCAR/musica/blob/25df47c4237c46b661064ec6a1c91db271690550/include/musica/configuration/emissions.hpp#L20-L21
2. https://github.com/NCAR/musica/blob/25df47c4237c46b661064ec6a1c91db271690550/include/musica/configuration/parse.hpp#L14-L15

However, both of these read the same file and produce something usable for either [micm](https://github.com/NCAR/musica/blob/25df47c4237c46b661064ec6a1c91db271690550/include/musica/configuration/parse.hpp#L16) or [miem](https://github.com/NCAR/musica/blob/25df47c4237c46b661064ec6a1c91db271690550/include/musica/configuration/emissions.hpp#L22). Instead, let's consolidate down to one read path that produces a mechanism and then we pass that mechanism to either miem/micm

We should consolidate this

**Acceptance criteria**
- Only one pathway exists to read a configuration file path or string and it returns a mechanism
  - Meaning, these functions should be removed 
    - https://github.com/NCAR/musica/blob/25df47c4237c46b661064ec6a1c91db271690550/src/configuration/parse.cpp#L19-L27
    - https://github.com/NCAR/musica/blob/25df47c4237c46b661064ec6a1c91db271690550/src/configuration/emissions.cpp#L148-L155
  - and all parsing should flow through https://github.com/NCAR/musica/blob/main/src/configuration/read_mechanism.cpp
- A function exist which takes a mechanism and can configure micm with a [`Chemsitry`](https://github.com/NCAR/musica/blob/main/include/musica/configuration/chemistry.hpp) object
- A function exists which takes a mechanism and can configure miem with an [`Emissions`](https://github.com/NCAR/musica/blob/25df47c4237c46b661064ec6a1c91db271690550/include/musica/configuration/emissions.hpp#L14-L18) object
- parsing a file can still work even if miem/micm are not enabled in a build
  - if possible, do this without ifdefs, but it may be unavoidable
  - might already be true in [`read_mechanism.cpp`](https://github.com/NCAR/musica/blob/main/src/configuration/read_mechanism.cpp), but confirm
- One parse can feed many converters, and there is a test proving it
  - `ReadMechanism` once, then `ConvertMechanism` **and** `ConvertEmissions` on the same mechanism, with no second parse.
- [`ConvertEmissions` throws `MissingEmissionsSection`](https://github.com/NCAR/musica/blob/25df47c4237c46b661064ec6a1c91db271690550/src/configuration/emissions.cpp#L120) when a mechanism has no emissions. Missing emissions should no longer be an error

**Ideas**
- Consider renaming `ConvertMechanism` to `ConvertChemistry` (or `ConvertToChemistry`) to mirror `ConvertEmissions`

smeechncar:
## Summary
- Remove the redundant `ReadConfiguration`/`ReadConfigurationFromString` (parse.cpp) and `ReadEmissionsConfiguration`/`ReadEmissionsConfigurationFromString` (emissions.cpp) wrappers, which each independently re-parsed a config file/string. All callers now call `ReadMechanism`/`ReadMechanismFromString` once and pass the resulting `Mechanism` to whichever converter(s) they need.
- Rename `ConvertMechanism` to `ConvertChemistry` to mirror `ConvertEmissions`, per the issue's suggestion.
- `ConvertEmissions` no longer throws when a mechanism has no emissions section, it returns an empty `Emissions` instead. Removed the now-unused `MissingEmissionsSection` error code.
- Added `src/test/unit/configuration/`: direct tests for `ReadMechanism`/`ReadMechanismFromString` (built regardless of MICM/MIEM), and a test proving one parse can feed both `ConvertChemistry` and `ConvertEmissions` with no second parse (gated on both MICM and MIEM being enabled).

Closes #932

## Test plan
- [x] Built and ran full test suite with `MUSICA_ENABLE_MIEM=ON` (Clang 17), 10/10 tests pass, including the new `read_mechanism` and `one_parse_many_converters` tests
- [x] Built and ran full test suite with default flags (`MUSICA_ENABLE_MIEM=OFF`), 6/6 tests pass, confirming `emissions.cpp` and the MIEM-only test are correctly excluded and parsing still works without MICM/MIEM ifdefs
- [x] Verified no remaining references to the removed functions/error code anywhere in the repo


K20shores (approved):
This matches what I was thinking. Maybe as the usage of miam/miem matures we will have to do some more refactoring, but I think a big win here is a single place to read rather than more than one

boulderdaze (approved):

PR #941, opened 2026-07-14
version 0.16.1

K20shores:
Update the version and enable testing of miam notebooks so we can release miam in python to get the binder link working.

We are only releasing the python library, so we won't do any of the other releases. This is one downside of a mono repository where we try to version all packages the same

notebook tests were added in #919, but that same PR disabled miam
miam was re-enabled in #930


boulderdaze (changes requested):
Please check updating version in citation, package.json, etc

K20shores:
> Please check updating version in citation, package.json, etc

I'm not updating javascript/julia interfaces so I didn't bump those versions, but I just did bump the citation version

boulderdaze (approved):

Issue #949, opened 2026-07-14
Build miem musica wrappers


K20shores:
Similar to micm, tuvx, we should make a musica wrapper around miem types to provide convenience functions. This will allow C/Python/Fortran wrappers to target the same interface

**Acceptance criteria**
- Add `musica::EmissionsModel`, or another name if you have a better one
  - wraps `miem::EmissionsBuilder`, `miem::Emissions`. Input is the `musica::Emissions` produced
  by `ConvertEmissions(mechanism)` and grid dimemsions (`n_cells`, `n_vert_levels`). 
    - Methods
      - `Run(epoch_seconds, dt)` returning/holding state; accessors 
      - `NumSpecies()`
      - `SpeciesNames()`
      - `SurfaceFlux(cell, species)`
    - Add a convenience `FromMechanism(mechanism, n_cells, n_vert_levels)` that internally
  calls `ConvertEmissions`
- Migrate and rename `src/test/unit/miem/end_to_end*.cpp` onto the musica emissions interface.

**Ideas**
- follow conventions in micm/tuvx c++ classes

Issue #22, opened 2023-05-31
Move MUSICA CCPP wrapper to github.com/NCAR/atmospheric_physics


K20shores:

K20shores:
https://github.com/NCAR/atmospheric_physics/pull/63

PR #960, opened 2026-07-16
Build miem python bindings

### LINKED ISSUE #942: Build miem python bindings
miem is now included in our build and has c++ tests, but to demonstrate it we need to write a python interface and demonstrate its usage in a box model. This depends on #947

Target Python API:
```python
mechanism = musica.mechanism_configuration.parse("config.yaml")  
emissions = musica.miem.Emissions(mechanism=mechanism, n_cells=N, n_vert_levels=1)
flux  = emissions.run(epoch_seconds, dt)
names = emissions.species_names
```

**Acceptance criteria**
- miem python bindings are made to the C++ musica classes
- tests are added which
  - verify
- an in-code mechanism is created in a notebook to demonstrate how miem can be used to drive emissions in a box model
- Add `python/bindings/miem/miem.cpp` with `bind_miem(py::module_&)` 
  - proposed methods
   - `_create_emissions_from_mechanism(mechanism, n_cells, n_vert_levels)`
   - `run(...)`
   - `num_species`
   - `species_names`
   - and surface-flux access as zero-copy numpy arrays where practical
- Register the submodule: add `bind_miem` to `python/bindings/common.hpp`
 - create and call it in `python/bindings/common.cpp`; add to `python/bindings/CMakeLists.txt`
- Add a pure-Python wrapper package `python/musica/miem/` with an `Emissions` class
  taking `mechanism=` and grid deimsnions using `CppWrapper`(`python/musica/_base.py`) and the `self._cpp` composition pattern; `_unwrap(mechanism)`
- Tests under `python/test/`: file-based configure and `run`, and
  an in-code emissions-object construction path.

**Ideas**
- consider creating another tutorial based on the chapman box model to include NOx chemistry and drive the emissions with miem

smeechncar:
## Summary
- Adds `musica.miem.Emissions`, a Python wrapper around `musica::EmissionsModel` — `python/bindings/miem/miem.cpp` (pybind11 binding, registered in `common.hpp`/`common.cpp`/`CMakeLists.txt`) plus the pure-Python wrapper `python/musica/miem/emissions.py`, matching the issue's proposed API (`Emissions(mechanism=..., n_cells=..., n_vert_levels=...)`, `run(epoch_seconds, dt)`, `num_species`, `species_names`, zero-copy numpy surface-flux access).
- Fixes a packaging gap: `python/bindings/CMakeLists.txt` wasn't installing `libmiem.so` into the wheel/editable-install dir alongside the other backend libraries — invisible before since #949 never built the Python extension with MIEM on.
- Two tutorial scripts under `python/musica/examples/` (plus a notebook, `tutorials/17. miem_nox_box_model.ipynb`) demonstrating miem driving real NOx chemistry in a single-box TS1 + MICM + TUV-x model:
  - `miem_nox_box_model.py` — synthetic single-cell, single-snapshot inventory.
  - `miem_nox_box_model_real_fixture.py` — the real committed CAMS-GLOB-ANT NOx fixture (`configs/miem/x1.163842_2024_nox_subset.nc`, 12 monthly 2024 snapshots × 4097 cells), exercising the real one-to-many species split (`nox_anth_sum` → NO 0.9 / NO2 0.1) and linear temporal interpolation across real monthly brackets.
- Fixes a real, pre-existing bug found while verifying the above with a working Fortran toolchain (Intel `ifx`) and TUV-x/CARMA enabled: `src/tuvx/interface.F90`/`src/carma/interface.F90`'s version-string getters passed a pointer reconstructed via `c_f_pointer` to `DEALLOCATE`, which `ifx` correctly rejects (gfortran silently tolerated it) — segfaulted `import musica` on any Intel-toolchain build with TUV-x on, unrelated to MIEM. Backed the version string with a module-level `SAVE`d `allocatable, target` buffer instead.
- Minor fix in the tutorial: `np.array([t.timestamp() for t in sim_times], dtype="datetime64[s]")` rejects raw Python floats — cast to `int` first.

Closes #942

## Test plan
- [x] `pytest python/test/unit/miem/test_emissions.py -v` — both required paths from the issue's acceptance criteria pass: `TestEmissionsFileBasedConfigureAndRun` (parses a mechanism YAML from disk) and `TestEmissionsInCodeConstruction` (builds `EmissionsConfig`/`Mechanism` directly in Python), each asserting the NOx→NO/NO2 9:1 split.
- [x] `python python/musica/examples/miem_nox_box_model.py` — runs to completion, no solver-not-converged warnings; output matches expected NOx/O3 photochemistry (NO rises, O3 titrates down, NO2 grows then plateaus).
- [x] `python python/musica/examples/miem_nox_box_model_real_fixture.py` — runs to completion against the real fixture; NO/NO2 flux splits exactly 9:1 at the real July-2024 magnitude, chemistry response matches the synthetic version's shape scaled to the higher real flux.
- [x] Verified `import musica` no longer segfaults with `MUSICA_ENABLE_TUVX=ON`/`MUSICA_ENABLE_CARMA=ON` on Derecho's Intel toolchain (`icx`/`icpx`/`ifx`), where it previously did.

## Note for reviewers
If you build with `MUSICA_ENABLE_MIEM=ON` and hit a segfault on `import musica` with TUV-x/CARMA enabled on an Intel-toolchain (`ifx`) system, make sure you're on this branch's tip — the `interface.F90` fix above addresses exactly that.


boulderdaze:
Hi @smeechncar, just a heads-up that my PR #963 should fix the notebook failure in the CI action. The notebooks are currently using the old APIs, and I've included the updates in the PR.

smeechncar:
Thanks @boulderdaze yeah I figured that was a problem on the main, pre-existing this branch. Currently just trying to work through these Windows build errors which seem to be some sort of netcdf build issue. Each test takes a really long time to complete though which why there is a delay on this PR.

K20shores:
> Thanks @boulderdaze yeah I figured that was a problem on the main, pre-existing this branch. Currently just trying to work through these Windows build errors which seem to be some sort of netcdf build issue. Each test takes a really long time to complete though which why there is a delay on this PR.

netcdf should already be available. We supply it for tuvx. Maybe miam is looking for cmake names we aren't exporting in musica's cmake?

smeechncar:
@K20shores I don't think I can sum it up better than Claude can at the moment. This one is leaving me pretty confused:

  Windows wheel-build saga on PR #960 — status:

  1. Root cause: miem's own cmake/FindnetCDF.cmake only checks nc-config/NETCDF_DIR/NETCDF_ROOT, so on the Windows CI runners (MSYS2/mingw,
  netCDF installed via pacman) it never finds the existing install and falls back to FetchContent-building netcdf-c v4.9.2 from source —
  which fails to compile under MinGW-UCRT (dpathmgr.c:704: passing argument 2 of '_wstat64' from incompatible pointer type). This is the
  first time miem's CMake has ever run on Windows CI (MIEM was only just enabled there).
  2. Applied a short-term hint on the musica side (NETCDF_ROOT=$MINGW_PREFIX in prepare_build_environment_windows.sh) to unblock, but the
  real fix belongs in miem's FindnetCDF.cmake.
  3. Currently testing the real fix by pointing MIEM_GIT_TAG at miem PR #22 (Windows netCDF discovery fix) — confirmed this resolved the
  netCDF issue: both Windows wheel builds got past compile/link cleanly this run.
  4. That surfaced a second, unrelated bug: test_emissions.py's file-based test embeds a raw Windows tmp_path inside a double-quoted YAML
  string, and yaml-cpp choked on backslash sequences like \msys64 as an invalid escape char. Fixed (868953c) by using .as_posix() for the
  embedded path.
  5. Net: waiting on this latest CI run. If green, the only remaining loose end is that #960 currently depends on unmerged miem PR #22 — that
  needs to land in miem and get re-pinned to a real commit before #960 can merge.

K20shores:
> @K20shores I don't think I can sum it up better than Claude can at the moment. This one is leaving me pretty confused:
> 
> Windows wheel-build saga on PR #960 — status:
> 
> 1. Root cause: miem's own cmake/FindnetCDF.cmake only checks nc-config/NETCDF_DIR/NETCDF_ROOT, so on the Windows CI runners (MSYS2/mingw,
>    netCDF installed via pacman) it never finds the existing install and falls back to FetchContent-building netcdf-c v4.9.2 from source —
>    which fails to compile under MinGW-UCRT (dpathmgr.c:704: passing argument 2 of '_wstat64' from incompatible pointer type). This is the
>    first time miem's CMake has ever run on Windows CI (MIEM was only just enabled there).
> 2. Applied a short-term hint on the musica side (NETCDF_ROOT=$MINGW_PREFIX in prepare_build_environment_windows.sh) to unblock, but the
>    real fix belongs in miem's FindnetCDF.cmake.
> 3. Currently testing the real fix by pointing MIEM_GIT_TAG at miem PR [Move MUSICA CCPP wrapper to github.com/NCAR/atmospheric_physics #22](https://github.com/NCAR/musica/issues/22) (Windows netCDF discovery fix) — confirmed this resolved the
>    netCDF issue: both Windows wheel builds got past compile/link cleanly this run.
> 4. That surfaced a second, unrelated bug: test_emissions.py's file-based test embeds a raw Windows tmp_path inside a double-quoted YAML
>    string, and yaml-cpp choked on backslash sequences like \msys64 as an invalid escape char. Fixed ([868953c](https://github.com/NCAR/musica/commit/868953ce76829c9f90f30e0c275ad43edce67f08)) by using .as_posix() for the
>    embedded path.
> 5. Net: waiting on this latest CI run. If green, the only remaining loose end is that [Build miem python bindings #960](https://github.com/NCAR/musica/pull/960) currently depends on unmerged miem PR [Move MUSICA CCPP wrapper to github.com/NCAR/atmospheric_physics #22](https://github.com/NCAR/musica/issues/22) — that
>    needs to land in miem and get re-pinned to a real commit before [Build miem python bindings #960](https://github.com/NCAR/musica/pull/960) can merge.

Sooo #22 is from 3 years ago and not something we should be targeting. That is entirely unrelated to this work

smeechncar:
@K20shores Oops Github's auto linker filled in the wrong link for "miem PR #22" which is actually supposed to be this: https://github.com/NCAR/miem/pull/22

boulderdaze:
> Thanks @boulderdaze yeah I figured that was a problem on the main, pre-existing this branch. Currently just trying to work through these Windows build errors which seem to be some sort of netcdf build issue. Each test takes a really long time to complete though which why there is a delay on this PR.

Could you pull the main and merge it into your branch, so we can see how the CI tests are going?

boulderdaze:
> > Thanks @boulderdaze yeah I figured that was a problem on the main, pre-existing this branch. Currently just trying to work through these Windows build errors which seem to be some sort of netcdf build issue. Each test takes a really long time to complete though which why there is a delay on this PR.
> 
> Could you pull the main and merge it into your branch, so we can see how the CI tests are going?

Thanks! Just add me as a reviewer when it's ready to review

smeechncar:
@K20shores @boulderdaze status update on the Windows wheel-build saga, and I'd like your input on how to proceed.

Since the netCDF discovery fix (`4b017bb`), a new issue showed up on **both** Windows wheel jobs: `pytest` finishes and reports all tests passing, but the process then hangs for the rest of the job (~80 min of silence) before the CI timeout kills it. This looks like a native interpreter-shutdown hang inside the compiled `_musica` extension, not a test failure — same general category as the earlier TUV-x/CARMA `c_f_pointer`/`DEALLOCATE` bug we fixed on this branch, just surfacing during teardown instead of at import.

I've tried two independent in-process fixes to skip the hanging teardown path:
1. `os._exit()` instead of `sys.exit()` (`6be9fb8`) — no change, still hung (Windows' `ExitProcess()` still delivers `DLL_PROCESS_DETACH` even via `os._exit()`).
2. `TerminateProcess()` via ctypes, which skips `DLL_PROCESS_DETACH` entirely (`86cbeac`) — this **fixed** `windows-11-arm ARM64` (passing for the first time), but `windows-latest AMD64` hung instead. Re-ran that job alone and it hung again the same way, so it's not just one-off flakiness.

Every other check on the PR is green. At this point I've spent a few days iterating on in-process fixes and don't want to keep guessing solo. Two ways I see to move forward:
- **(a)** Wrap the whole Windows test-command in an external OS-level watchdog (launch the process, wait for a completion sentinel, force-kill from outside via `taskkill /F`) — same idea as `TerminateProcess()` but invoked externally instead of from within the possibly-stuck process.
- **(b)** Mark this one Windows wheel test-command as allowed-to-fail for now, with a tracking issue, so this PR isn't blocked on a pre-existing-category-of-bug that's orthogonal to the actual MIEM Python bindings work.

Which of these (or something else) would you prefer? Happy to implement either.

K20shores:
Ah, I should have seen this coming. @smeechncar this is a known issue of netcdf on windows. I submitted a [pr to netcdf to fix this](https://github.com/Unidata/netcdf-c/pull/3400). Basically, the AWS SDK bundled with netcdf on windows never closes it thread on process shutdown and the whole process hangs. AWS is included to read data in S3 and netcdf was always initializing AWS, even if no data from S3 would ever be read. 

Disable the miem tests on windows by overriding the test command in pyproject.toml for windows. That's what we do for tuvx

smeechncar:
Now that MIEM's PR (https://github.com/NCAR/miem/pull/22) is merged, the MIEM_GIT_TAG in MUSICA has been re-pinned to it in the latest push here. Should be ready for review now that we are skipping miem tests on Windows wheel builds.

python/musica/examples/miem_nox_box_model.py
```diff
@@ -0,0 +1,273 @@
...
+        while current_time < simulation_length:
+            flux = emissions.run(sim_time.timestamp(), time_step)
+            no_index = list(emissions.species_names).index("NO")
+            no_surface_flux = flux[no_index, 0]  # kg m-2 s-1, single cell
+
+            air_density = state.get_conditions()["air_density"][0]  # mol m-3
+            delta_no_mol_m3 = no_surface_flux * time_step / (BOX_HEIGHT_M * NO_MOLECULAR_WEIGHT)
+
+            current_concentrations = state.get_concentrations()
+            current_concentrations["NO"] = [current_concentrations["NO"][0] + delta_no_mol_m3]
```
K20shores: You are overwriting all grid cells, with the zeroth grid cell's concentration and delta. I think this might be what you meant to do

```suggestion
            current_concentrations["NO"] = [concentration + delta_no_mol_m3 for concentration in current_concentrations["NO"]]
```

python/musica/examples/miem_nox_box_model_real_fixture.py
```diff
@@ -0,0 +1,278 @@
...
+    last_printed_percent = -5
+
+    while current_time < simulation_length:
+        flux = emissions.run(sim_time.timestamp(), time_step)
+
+        current_concentrations = state.get_concentrations()
+        for name in species_names:
+            surface_flux = flux[species_indices[name], CELL_INDEX]  # kg m-2 s-1
+            delta_mol_m3 = surface_flux * time_step / (BOX_HEIGHT_M * MOLECULAR_WEIGHTS[name])
+            current_concentrations[name] = [current_concentrations[name][0] + delta_mol_m3]
```
K20shores: same comment here

python/tools/run_pytest_with_watchdog.py
K20shores, line 1: remove this

src/carma/interface.F90
K20shores, line 1: why are we touching these fortran interfaces? If they are wrong, please file an issue. Either way, please revert these changes so that this is only a miem PR

tutorials/19. miem_nox_box_model.ipynb
K20shores, line 1: you'll need to bump the number. I think main now already has a 17 since you submitted this PR. Also please check where we list our tutorials in our docs. That list probably needs to be updated with the new tutorials as well as yours

K20shores (changes requested):

src/carma/interface.F90
smeechncar: This popped up when I was using Intel ifx on Derecho. Intel seems more strict than gfortran when importing musica with a "forrtl: severe (173): A pointer passed to DEALLOCATE points to an object that cannot be deallocated" message from both internal_get_tuvx_version and internal_free_tuvx_version.  
I can revert this and change my build on Derecho to gfortran for now. I think its a good idea to create an issue for it.

python/musica/examples/miem_nox_box_model.py
smeechncar: yes thanks change applied

python/tools/run_pytest_with_watchdog.py
smeechncar: removed

boulderdaze (approved):

tutorials/19. miem_nox_box_model.ipynb
smeechncar: merged main back in to this branch, set our chapter to the next chapter 19.

K20shores (changes requested):
Between `python/musica/examples/miem_nox_box_model.py` and `python/musica/examples/miem_nox_box_model_real_fixture.py`, we show examples of creating emissions. However, both of those are done after creating an in-code mechanism specifically for emissions. I'd also like to see how we could read one configuration file that has a full mechanism *and* emissions, and get back a solver and emissions calculator. 

If you think this would best be done as a refactor in a follow-on PR, that would also be fine.

K20shores (changes requested):
Please clean up the language in the notebook. It reads like it was written by an LLM. I know it was written by an LLM, but there is a lot of extraneous detail that does not belong in the notebook.

python/musica/examples/miem_nox_box_model.py
```diff
@@ -0,0 +1,273 @@
...
+        current_time = 0
+        last_printed_percent = -5
+
+        while current_time < simulation_length:
+            flux = emissions.run(sim_time.timestamp(), time_step)
+            no_index = list(emissions.species_names).index("NO")
+            no_surface_flux = flux[no_index, 0]  # kg m-2 s-1, single cell
+
+            air_density = state.get_conditions()["air_density"][0]  # mol m-3
+            delta_no_mol_m3 = no_surface_flux * time_step / (BOX_HEIGHT_M * NO_MOLECULAR_WEIGHT)
```
K20shores: @smeechncar you seem to be using the flux to convert to an instantaneous concentration. Instead, we should be using emissions reactions for each NOx species we care about. Then we would use miem to set the rates of emissions.

K20shores (changes requested):

python/musica/examples/miem_nox_box_model.py
smeechncar: OK I see what happened there. We're now using the same pattern in sulfate_box_model.py and instead of converting MIEM's flux into a concentration change, we convert it into a rate. In this case, since the time delta is small, the results look almost identical to the previous method but the change seems to be working.

K20shores (approved):

smeechncar:
> Between `python/musica/examples/miem_nox_box_model.py` and `python/musica/examples/miem_nox_box_model_real_fixture.py`, we show examples of creating emissions. However, both of those are done after creating an in-code mechanism specifically for emissions. I'd also like to see how we could read one configuration file that has a full mechanism _and_ emissions, and get back a solver and emissions calculator.
> 
> If you think this would best be done as a refactor in a follow-on PR, that would also be fine.

I think moving this to another PR is the right move. 

I've also removed some of the fluff in the notebook that didn't read very well and left just the technical substance.

PR #964, opened 2026-07-17
Update GitHub Actions

zmoon:
and add Dependabot config for future automatic PRs


K20shores (approved):
Oh, neat. I only updated these by hand when I noticed something was out of date. I should have known that there was an automated way to do this. Thanks!

PR #970, opened 2026-07-18
Aerosol Intro Tutorial and debugging

mattldawson:
This PR adds an "Intro to Aerosol Modeling" tutorial to walk through the conceptual approach to multi-phase modeling used in MUSICA. It also includes some light clean up of the other aerosol and cloud related notebooks and the minor updates below.

# Minor Updates
- Uses latest MIAM and MechanismConfiguration tags with renamed Henry's Law components
- Renamed `*rate_constants` as `*rate_constant` when only a single rate constant is expected.
- Fixes an issue in the Python API that prevented physical properties associated with species from being included in certain processes.
- Allows `PhaseSpecies` to be created based on an existing `Species` instead of only by a name string
- Adds a Copilot-written set of integration tests for individual MIAM components


python/test/integration/test_miam_components.py
```diff
@@ -0,0 +1,1124 @@
...
+# DissolvedEquilibrium, LinearConstraint).  Every test is parametrized over
+# all three aerosol representations (UniformSection, SingleMomentMode,
+# TwoMomentMode) to catch representation-specific bugs.
+#
+# Species, rate constants, and initial conditions are deliberately made-up
+# and chosen to make the analytical solution simple to compute.
+#
+# Analytical solutions used
+# ─────────────────────────
+# DissolvedReaction  foo → bar, rate = k·[S]/([S]+ε)·[foo]
```
K20shores: I prefer if we remove the unicode characters and replace them with ascii

K20shores (changes requested):

boulderdaze (approved):

python/test/integration/test_miam_components.py
mattldawson: I think they should be removed now

PR #980, opened 2026-07-22
Update ts1 reaction types

### LINKED ISSUE #768: Add TS1 fortran box model example
In #753 and #766 we created a python column model that uses TS1 with photolysis rates driven by tuvx.

**Acceptance criteria**
- Add a fortran example which does the same thing
- This is added as a tutorial chapter

**Ideas**
- Look at atmospheric physics to see how to use the mappings file
- Potentially create a mappings file from the `__CAM` part of the configuration file, remove that part and update the python code

K20shores:
Closes #768 

- updates the TS1 configuration to rename the photolysis rate reactions from user defined to photolysis
- updates the initial conditions and usage of them in python'
- creates a box model that runs TS1 in fortran as an example, as well as making another chapter


dwfncar (approved):
Looks good.

boulderdaze (approved):
How is this decision made to change `user_defined` to `photolysis`? This change is definitely more descriptive, but I'm curious how we can now use `photolysis` instead of `user_defined`. What does this look like on the MICM side?

K20shores:
> How is this decision made to change `user_defined` to `photolysis`? This change is definitely more descriptive, but I'm curious how we can now use `photolysis` instead of `user_defined`. What does this look like on the MICM side?

The v1 config was made by me translating the v0 config using the python musica cli awhile ago. [vo still has all of the rates listed as photolysis rates](https://github.com/NCAR/musica/blob/125230bf0ef32fcb914ba7a1d2a5fd5c049fcf2f/configs/v0/TS1/reactions.json#L76-L87). At the time, the tool translated all types that can have rates set to user defined types. I didn't care at the time, but I should have. Someone looking at the configuration would want to identify what is an emission reaction, a photolysis rate, etc.

The only difference is that the rate is set using `PHOTO.<reaction name>` instead of `USER.<reaction name>`. You can see that change in the initial conditions file for TS1

PR #988, opened 2026-07-26
Support photolysis reactions with no reactants

### LINKED ISSUE #986: v0 parser drops photolysis reactions with no reactants
`mechanism_configuration.parse()` silently drops v0 (CAMP) `PHOTOLYSIS` reactions whose `reactants` is empty (`{}`).

Music Box Interactive encodes **emissions** as reactant-less photolysis so they can carry an `irr__` product, e.g.:
```json
{
  "type": "PHOTOLYSIS",
  "__music_box_type": "EMISSION",
  "MUSICA name": "EMIS_IOLE",
  "reactants": {},
  "products": {"IOLE": {"yield": 1}, "irr__...": {"yield": 1}}
}
```

Repro (CB05 MBI config): the config has 39 photolysis reactions, 14 of them reactant-less; `parse()` returns only the 25 with reactants. Downstream this breaks music_box: the conditions CSV supplies rates for the dropped reactions, so the solver raises `User-defined rate parameter PHOTO.EMIS_NO not found in the mechanism.`

**Fix:** allow zero reactants in the v0 photolysis parser (mechanism_configuration `src/v0/photolysis_parser.cpp`) instead of dropping the reaction.

K20shores:
Pairs with **NCAR/MechanismConfiguration#312**, which makes `Photolysis.reactants` a list (0 or 1) so a reactant-less photolysis — an emission encoded as photolysis to carry an `irr__` product — is kept instead of dropped.

## Changes

- **Python `Photolysis` wrapper**: `reactants` getter/setter now use `_wrap_list`/`_unwrap_list` (mirroring `products`) instead of assuming exactly one reactant; at most one reactant is allowed, and `serialize()` emits an empty `reactants` list when there are none.
- **`cmake/dependencies.cmake`**: points `MECH_CONFIG_GIT_TAG` at the #312 branch so this builds against the list-valued type. **Must be updated to the merged commit of #312 before merging.**
- Adds a test that a reactant-less `Photolysis` is allowed and round-trips.

No C++ change is needed: `convert_user_defined` already consumes `reaction.reactants` via the vector overload (an empty list yields no reactants).

## Why

MBI encodes emissions as reactant-less photolysis; musica's v0 parser dropped them, so converted configs (e.g. CB05) failed to run with `PHOTO.EMIS_NO not found`. Keeping them as photolysis preserves the `PHOTO.<name>` rate-parameter label, so the conditions CSVs match with no remapping.

Closes #986 together with NCAR/MechanismConfiguration#312.

### DIFF
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -67,7 +67,9 @@ endif()
 
 if(MUSICA_BUILD_C_CXX_INTERFACE AND NOT MUSICA_USE_PREBUILT)
   set_git_default(MECH_CONFIG_GIT_REPOSITORY https://github.com/NCAR/MechanismConfiguration.git)
-  set_git_default(MECH_CONFIG_GIT_TAG 2fb23128b3828e84c0bb73f87a678af3a8c3786c)
+  # TODO: update to the merged commit of NCAR/MechanismConfiguration#312 before merging;
+  # this points at that PR's branch so photolysis reactants are list-valued here.
+  set_git_default(MECH_CONFIG_GIT_TAG 95a828f6ac1fd8eec60823c39c828f81001e7a3d)
 
   FetchContent_Declare(mechanism_configuration
       GIT_REPOSITORY ${MECH_CONFIG_GIT_REPOSITORY}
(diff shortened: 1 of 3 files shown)


K20shores:
Folded into #987 (update_deps), which already bumps the mechanism_configuration tag to the merged #312. Added the Python photolysis wrapper change there so the version bump and the paired Python change ship together.

PR #990, opened 2026-07-28
Build miem C/Fortran interface

### LINKED ISSUE #943: Build miem C/fortran interface
Expose a C ABI and write fortran bindings to them to expose miem object, as we do for micm/tuvx

**Acceptance criteria**

- Add `include/musica/miem/emissions_c_interface.hpp` and
  `src/miem/emissions_c_interface.cpp` (`extern "C"`) on top of the musica emissions interface from #949.
  - proposed methods
    - `CreateEmissions(Mechanism* mechanism, int n_cells, int n_vert_levels, Error*)`
    - `DeleteEmissions`
    - `EmissionsRun(epoch, dt,...)`
    - accessors: `GetNumSpecies`, `GetSpeciesOrdering`, `GetSurfaceFluxPointer` / strides). 
- Fortran API
  - Add `fortran/miem/emissions.F90` (`module musica_emissions`): `bind(C)` interfaces
  - an `emissions_t` derived type holding `type(c_ptr)` with type-bound `run` / getters and a `final` calling `DeleteEmissions`.
- Add C/Fortran tests mirroring the MICM Fortran tests
- Add tests showing how to use miem in a box model. If the python tests in #942 added a chapman nox box model, replicate that here

**Ideas**
- mirror the micm/tuvx interfaces very closesly

smeechncar:
## Summary

- Adds an `extern "C"` interface for `musica::EmissionsModel` (`include/musica/miem/emissions_c_interface.hpp` / `src/miem/emissions_c_interface.cpp`): `CreateEmissions`, `DeleteEmissions`, `EmissionsRun`, `GetNumSpecies`, `GetEmissionsSpeciesOrdering`, `GetSurfaceFluxPointer`/`GetSurfaceFluxStrides` — mirroring the existing MICM/TUV-x C interfaces closely.
- Adds a Fortran `musica_emissions` module (`fortran/miem/emissions.F90`): `mechanism_t`/`emissions_t`, `bind(C)` interfaces, type-bound `run`/getters, `final` destructors — mirroring `musica_state`'s shape.
- Since `CreateEmissions` takes a `Mechanism*` and no C-visible `Mechanism` handle existed anywhere in musica before this, also adds `ReadMechanismC`/`ReadMechanismFromStringC`/`DeleteMechanism` (`include/musica/configuration/read_mechanism_c_interface.hpp` / `src/configuration/read_mechanism_c_interface.cpp`) to expose `mechanism_configuration::Mechanism` as an opaque C pointer for the first time.
- Adds `EmissionsModel::NumCells()`/`SurfaceFluxData()` to back the pointer/strides accessors (a normalized `double` cache refreshed on every `Run()`, independent of `miem::Real`'s build-time precision).
- Adds C++ and Fortran tests: `src/test/unit/configuration/read_mechanism_c_interface.cpp`, `src/test/unit/miem/emissions_c_api.cpp`, `fortran/test/integration/test_emissions_api.F90`.
- Adds two Fortran box-model examples mirroring the Python tutorials from #942: `fortran/examples/miem_nox_box_model.F90` (synthetic single-cell NO inventory) and `fortran/examples/miem_nox_box_model_real_fixture.F90` (real committed `x1.163842_2024_nox_subset.nc` fixture, one-to-many NO/NO2 split, linear temporal interpolation). Both couple TS1 chemistry + TUV-x photolysis + MICM + a real miem emission flux, injected as `EMIS.NO`/`EMIS.NO2` MICM rate parameters — since `ts1.json` has no `EMISSION` reactions and there's no API to add one to an already-parsed mechanism, both examples splice one in via CMake at configure time (`ts1.json` itself is never modified).

Closes #943.

## Notes for reviewers

- `GetSpeciesOrdering` had to be renamed to `GetEmissionsSpeciesOrdering`: MICM's `state_c_interface.hpp` already defines an `extern "C" GetSpeciesOrdering(State*, ...)`, and two identically-named `extern "C"` symbols is a link failure, not a style choice.
- `CreateEmissions` takes `const Mechanism*` (`mechanism_configuration::Mechanism`) directly, matching the issue's proposed signature literally, rather than a config-path/string convenience overload like `CreateMicm`'s three-overload pattern. Fortran's `mechanism_t` therefore only has a from-path constructor — a from-string variant can't coexist under the same generic `interface mechanism_t` (identical `character(len=*)+error_t` dummy-argument shape is not overload-resolvable in Fortran), so `ReadMechanismFromStringC` is exercised by a C++ test but has no Fortran caller yet.
- Verified with a full from-scratch build + `ctest` using the gfortran/g++/gcc toolchain (`-DMUSICA_BUILD_FORTRAN_INTERFACE=ON -DMUSICA_ENABLE_MICM=ON -DMUSICA_ENABLE_TUVX=ON -DMUSICA_ENABLE_MIEM=ON`), not Intel — matches the team's existing preference for gfortran on miem-adjacent Fortran work, since Intel's `ifx` has an unrelated, already-known segfault in `src/tuvx/interface.F90`/`src/carma/interface.F90` that this PR doesn't touch.

## Test plan

- [x] `ctest` passes fully (29/29), including all new targets: `read_mechanism_c_interface`, `miem_emissions_c_api`, `emissions_fortran_api`, `miem_nox_box_model`, `miem_nox_box_model_real_fixture`.
- [x] `miem_emissions_c_api`/`test_emissions_api.F90` assert the real fixture's NO:NO2 flux split is exactly 9:1, not just "some split happened".
- [x] Both box models assert the expected NOx photochemistry signature in-program (O3 net titration, NO2 growth, nonzero miem flux) — inspected the real CSV output directly as well, not just exit codes.


fortran/examples/CMakeLists.txt
```diff
@@ -62,3 +62,70 @@ add_test(NAME ts1_box_model
...
+        "type": "EMISSION",
+        "name": "NO",
+        "gas phase": "gas",
+        "products": [ { "species name": "NO", "coefficient": 1 } ]
+    },
+]=])
+  string(REPLACE "\"reactions\": [" "\"reactions\": [\n${NO_EMISSION_REACTION}"
+    TS1_WITH_NO_EMISSION "${TS1_JSON_CONTENTS}")
+  file(MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/configs/v1/ts1)
+  file(WRITE ${CMAKE_BINARY_DIR}/configs/v1/ts1/ts1_with_no_emission.json "${TS1_WITH_NO_EMISSION}")
```
K20shores: Not a fan of this. I'd much prefer a made up mechanism (meaning  a new config file) with enough reactions to show this off.

smeechncar: Replaced the CMake-time splice with a small standalone mechanism (`configs/v1/miem_nox/mechanism.json`) -- 3 species (NO/NO2/O3), the real `jno2` photolysis and `NO + O3 -> NO2` rate parameters pulled straight from `ts1.json`, plus the `EMISSION` reactions for NO/NO2. Both examples now just point at this file instead of a build-tree copy of `ts1.json` with reactions spliced in via `string(REPLACE ...)`.

Kept the TUV-x/TS1 photolysis coupling as-is (the existing aliasing pairs table already has a `jno2 -> jno2` entry we can reuse), so the box models still get a real, non-invented photolysis rate rather than a hardcoded one -- everything else about the time loop and miem coupling is unchanged.

Verified with a fresh build: both `miem_nox_box_model` and `miem_nox_box_model_real_fixture` still show the same NOx/O3 titration signature and 9:1 NO:NO2 flux split as before.

src/test/unit/miem/emissions_c_api.cpp
```diff
@@ -0,0 +1,256 @@
...
+// EmissionsRun call, matching MICM's GetOrderedConcentrationsPointer/state_t
+// re-fetch-after-every-solve convention. Note that in practice, for a fixed grid
+// size, std::vector::assign() commonly reuses the same backing allocation across
+// calls (confirmed here: the address does NOT change run to run), so this test
+// checks the documented contract's actual purpose -- that GetSurfaceFluxPointer
+// keeps returning correct, current data after repeated runs -- rather than
+// asserting on that implementation detail of std::vector, which is not part of
+// the API's contract and isn't guaranteed to hold under a different std::vector
+// implementation or a future change to EmissionsModel's caching.
+TEST_F(EmissionsCApiTestFixture, SurfaceFluxPointerStaysCorrectAcrossRepeatedRuns)
```
boulderdaze: Could you trim the description by focusing on what this test is actually testing? It's a bit long, and includes information that doesn't seem necessary. If API is already well documented, the test description can just focus on scenario being tested.
```yaml
 EmissionsRun call, matching MICM's GetOrderedConcentrationsPointer/state_t
// re-fetch-after-every-solve convention <- These kinds of details are unnecessary here.
```

src/test/unit/miem/emissions_c_api.cpp
```diff
@@ -0,0 +1,256 @@
+// Copyright (C) 2023-2026 University Corporation for Atmospheric Research
+// SPDX-License-Identifier: Apache-2.0
+//
+// Tests for the miem C interface (CreateEmissions/DeleteEmissions/EmissionsRun/
+// GetNumSpecies/GetEmissionsSpeciesOrdering/GetSurfaceFluxPointer/GetSurfaceFluxStrides),
+// added on top of the musica::EmissionsModel C++ wrapper from #949. Reuses the same
+// real committed NOx fixture and 9:1 NO:NO2 species-split config as
+// emissions_model_end_to_end_nox.cpp so the C API is checked against a known-correct
+// quantitative result, not just "it doesn't crash".
```
boulderdaze: One general comment for the PR,  could you review the comments and descriptions throughout your PR and trim them down if necessary?

We definitely don't need to include issue numbers like #949 anywhere.

fortran/examples/miem_nox_box_model.F90
```diff
@@ -0,0 +1,455 @@
...
+! so instead of splicing one in at build time, this uses a compact 3-species
+! (NO/NO2/O3) mechanism with the real jno2 photolysis and NO+O3 Arrhenius rate
+! parameters copied from ts1.json, plus the EMISSION reaction ts1.json lacks.
+!
+! Expected result: NO rises from the continuous emission, NO2 rises and
+! plateaus, and O3 is net titrated down (NO + O3 -> NO2 + O2) -- the same
+! photochemical signature the Python example produces.
+!
+program miem_nox_box_model
+  use iso_fortran_env, only: real64
```
boulderdaze, line 19: Looks like this test covers the integration test including tuvx, micm, and miem. What would be your next step? I'm curious.

boulderdaze (changes requested):

fortran/miem/emissions.F90
```diff
@@ -0,0 +1,276 @@
...
+      character(kind=c_char), intent(in)     :: config_path(*)
+      type(error_t_c),        intent(inout)  :: error
+      type(c_ptr)                            :: read_mechanism_c
+    end function read_mechanism_c
+
+    subroutine delete_mechanism_c(mechanism, error) bind(C, name="DeleteMechanism")
+      import c_ptr, error_t_c
+      type(c_ptr), value, intent(in)    :: mechanism
+      type(error_t_c),    intent(inout) :: error
+    end subroutine delete_mechanism_c
```
K20shores, line 30: @boulderdaze should we refactor micm so that it takes in a mechanism rather than a config file? Then we'd move these mechanism files to their own files and both miem and micm would only be useable after we've read a mechanism. It would open up the fortran side t if we made mechanisms in memory as well

src/test/unit/miem/emissions_c_api.cpp
K20shores: agreed

src/test/unit/miem/emissions_c_api.cpp
smeechncar: Good catch! I've cleaned up the comments in all of the files in this PR to match other file descriptors.

fortran/examples/miem_nox_box_model.F90
smeechncar: I would personally like to have tests that cover all of the CAMS and FINN variables. We only test BC and NOx so far from CAMS. The FINN files have been tested but not committed to the repo. I wonder if it would be a good idea to create a full workflow test that runs through all of the MVP variables, using real subsets of the CAMS and FINN files. This would be additional work on top of https://github.com/NCAR/musica/issues/991

fortran/miem/emissions.F90
boulderdaze: That makes sense to me. Would you like to write an issue?

boulderdaze (approved):

fortran/miem/emissions.F90
K20shores: Yes, #995

K20shores (approved):

K20shores (changes requested):
actually, are these changes tested in the CI? I don't see them. Perhaps I missed them

smeechncar:
Right, the CI leg that runs the full make test suite never sets -D MUSICA_ENABLE_MIEM=ON, and that option defaults to OFF in CMakeLists.txt. It does for the other modules though. I'm testing again now and will update when it is working.

K20shores (changes requested):
Please also enable the emissions tests in ubuntu, windows, mac

include/musica/miem/emissions_c_interface.hpp
```diff
@@ -0,0 +1,71 @@
...
+    /// @param error Error struct to indicate success or failure
+    /// @return Pointer to the flux buffer [kg m-2 s-1]
+    double* GetSurfaceFluxPointer(EmissionsModel* emissions, size_t* array_size, Error* error);
+
+    /// @brief Get the strides for indexing the surface flux buffer
+    /// @param emissions Pointer to EmissionsModel object
+    /// @param error Error struct to indicate success or failure
+    /// @param cell_stride Pointer to the stride across grid cells
+    /// @param species_stride Pointer to the stride across species
+    void GetSurfaceFluxStrides(EmissionsModel* emissions, Error* error, size_t* cell_stride, size_t* species_stride);
```
K20shores: Please move `Error* error` to the end for consistency with other APIs

K20shores (changes requested):

K20shores:
@smeechncar I keep forgetting that netcdf on windows doesn't work. You can turn MIEM off for windows fortran

K20shores (approved):

PR #1002, opened 2026-08-04
Add CAMS+FINN all-species MIEM MVP demo

smeechncar:
## Summary
- Extends the existing NOx real-fixture box model to prove every CAMS (anthropogenic + biogenic) and FINN (fire) MVP variable flows through the real pipeline: 14 real fixtures, one combined `cams_finn_all_species_emissions_config.yaml`, one demo script, one pytest suite.
- Each of the 14 sources gets a distinct category so flux for species fed by more than one source (e.g. MTERP by 5 sources) sums instead of one source overriding another.
- Fixes FINN's units: its fixture is in `molecules cm-2 s-1`, not `kg m-2 s-1` like the rest of the config -- confirmed with Forrest Lacey (MPAS-GOCART2G's biomass-burning convention). The "finn fire map" species-map scaling factors now convert it back to `kg m-2 s-1` per species before miem sums it with CAMS.
- Splits NOx into NO (90%) and NO2 (10%), the same ratio used by this repo's other NOx tests/demos, instead of mapping to NO2 only.
- NH3/CO/ISOP/MTERP/NO/NO2/SO2 already exist as gas-phase species in `ts1.json` and get a real MICM `Emission` reaction. BC/OC are aerosol mass with no gas-phase representation in this mechanism, so they get a flux-only panel instead.

Related: https://github.com/NCAR/miem/pull/24 (adds the real FINN fixture and documents its units on the miem side; not a build dependency -- this branch already carries its own copy of the fixture and an unrelated `MIEM_GIT_TAG` pin).

## Test plan
- [x] Full `python/test/unit/miem/` suite passes (35/35), including new regression tests confirming the FINN scaling factors apply as an exact linear multiply and the resulting flux is physically plausible
- [x] Ran the demo end-to-end against real fixtures; FINN-fed species now show physically sane flux/concentration behavior instead of the pre-fix unbounded buildup
- [x] Verified the NOx split produces an exact 9:1 flux ratio at the demo cell


smeechncar:
Added the Python test scripts and the subset files of real CAMS and FINN files to be used in a full pipeline MIEM workflow test. This ensures that all of the modules work correctly using all of the MVP variables.

python/musica/examples/miem_cams_finn_box_model_real_fixture.py
```diff
@@ -0,0 +1,269 @@
...
+    concentrations = [state.get_concentrations()]
+    flux_history = {name: [0.0] for name in species_names}
+    time_step = 30  # seconds
+    simulation_length = 3 * SECONDS_PER_HOUR
+    current_time = 0
+    last_printed_percent = -5
+
+    # emissions.run() opens the config's referenced fixtures by bare filename
+    # each call -- CWD must be emissions_config_dir for the whole loop, not
+    # just around building `emissions` above.
```
K20shores: this feels like something we should address, not create workarounds for. We should also consider caching several timesteps and not incurring a file open each run call

K20shores (commented):
Let me know what you think. Perhaps I'm misunderstanding, but it sounds like we baked in a poor API with a performance bottleneck, both of which are fairly easy to address, but perhaps I've misunderstood

configs/miem/cams_finn_all_species_emissions_config.yaml
```diff
@@ -0,0 +1,339 @@
...
+# individual sector variables -- "_sum" aggregates more sectors than just
+# those.
+#
+# BC/OC have no gas-phase representation in ts1.json, so they get real miem
+# flux but no MICM Emission reaction (flux-only, no concentration curve).
+version: 1.0.0
+species: []
+phases: []
+reactions: []
+emissions:
```
boulderdaze, line 26: The files with `.nc` extension are data files, and this contains configuration. 
The directory `configs/miem` contains a mix of yaml files and data files. I think it would be better to separate these into different directories to make the structure clearer and easier to maintain. What do you think?

python/musica/examples/miem_cams_finn_box_model_real_fixture.py
```diff
@@ -0,0 +1,269 @@
...
+
+    environmental_conditions = ussa1976.compute(z=np.array([vertical_edge * 1000]), variables=["t", "p"])
+    temperature = environmental_conditions['t'].values
+    pressure = environmental_conditions['p'].values
+
+    state.set_conditions(temperature, pressure)
+    state.set_concentrations(concentration_dict)
+    state.set_user_defined_rate_parameters(user_defined_dict)
+
+    emissions, emissions_config_dir = _get_emissions()
```
boulderdaze: It looks strange to me to have a public API function start with an underscore. Is there a reason it starts with an underscore?

python/musica/examples/miem_cams_finn_box_model_real_fixture.py
smeechncar: Dug into this. Two separate things going on here, and the actual bottleneck is smaller than the code around it suggests.

**The API isn't reopening the file every call.** `Emissions.run()` → `EmissionsModel::Run` → `miem::Emissions::Run` → each source's `OfflineEmissionSource::Update()`. `Update()` only reloads from disk when the current sim time falls outside the cached two-slice time bracket (`source_offline.cpp:194-203`, `TemporalInterpolator`); if the bracket still covers the timestep, it's an in-memory linear blend with no I/O. For this demo, CAMS is monthly-resolution and the FINN fixture is daily, so a 3-hour/30s-step run stays inside one bracket per source the whole time — that's ~14 file opens total (one per source), not 360.

It would become a real recurring cost against a finer-resolution product (e.g. full-resolution hourly FINN, vs. the daily fixture here) or a run spanning many bracket boundaries, since `LoadBrackets` always reopens from scratch on a miss with no "already open on this path" fast path (`source_offline.cpp` / `uptempo_reader.cpp`). That's worth hardening upstream in NCAR/miem, but it's a bracket-boundary-miss cost, not a per-timestep one.

**The actual bug is the `directory: ""` in the YAML, which is separate from the above.** `Inventory.directory` is already wired into path resolution (`src/configuration/emissions.cpp:101` — `std::filesystem::path(inventory.directory) / inventory.file_pattern`), it's just unset for all 14 inventories in `cams_finn_all_species_emissions_config.yaml`, which is why the resolved path is a bare filename and the example needs the `os.chdir` workaround. Setting `directory` to the fixtures' path removes the chdir hack entirely — no core change needed. I'll make that fix in this PR.

For the reopen-on-bracket-miss case, I'll open a follow-up issue against NCAR/miem rather than scoping it into this PR.

configs/miem/cams_finn_all_species_emissions_config.yaml
smeechncar: Good point. Moved the 13 CAMS+FINN fixtures this PR adds into `configs/miem/data/`, mirroring the existing `configs/tuvx/data` precedent, and set `directory: "data"` on the corresponding inventories in the yaml. `x1.163842_2024_nox_subset.nc` stays at the top level since it's shared with `nox_emissions_config.yaml` and `nox_real_fixture_emissions_config.yaml`, which are outside this PR's scope — moving it would mean touching two configs and a couple of Fortran targets that aren't part of this change.

This also let me fully resolve the chdir concern from the other thread: `_get_emissions()` and the test's `mechanism` fixture now rewrite each inventory's `directory` to an absolute path right after `parse()` (new `resolve_inventory_directories()` helper in `miem_cams_finn_shared.py`), so path resolution no longer depends on the process's CWD at all — the `os.chdir`/`monkeypatch.chdir` calls are gone.

python/musica/examples/miem_cams_finn_box_model_real_fixture.py
smeechncar: This matches the existing convention in the sibling scripts `miem_nox_box_model.py` and `miem_nox_box_model_real_fixture.py`, which both name their equivalent helper `_get_emissions` too, plus `_write_synthetic_no_inventory` in the first one. These are standalone example scripts, not a library module with an exported surface, so the leading underscore marks the function as local to the script rather than a public API function.

Happy to rename all three (across this PR and the two existing scripts) if you'd rather drop the underscore convention for these examples — let me know which way you'd like it.

boulderdaze: Ah, I understand. I missed the details. Thank you for the explanation!

boulderdaze (approved):
Thank you for addressing the comments. I don't have any additional comments besides the one that's currently being discussed.

K20shores (approved):

PR #1033, opened 2026-08-21
Test that a missed radiation field update reuses the last field

### LINKED ISSUE #1022: Radiation field updater: Python bindings
Part of #1018 (host-supplied radiation field). Third of four staged sub-issues, stacked on #1021 (Fortran interface bridge).

**Acceptance criteria**
- `python/bindings/tuvx/radiation_field_updater.cpp` binds `musica::RadiationFieldUpdater::Update`, accepting numpy arrays for the 3 required + 3 optional 2D arrays
- `python/musica/tuvx/radiation_field_updater.py` wraps it as a `RadiationFieldUpdater` class with `update(direct_actinic_flux, upward_actinic_flux, downward_actinic_flux, direct_irradiance=None, upward_irradiance=None, downward_irradiance=None)`, each shaped `(n_vertical_interfaces, n_wavelength_bins)`
- `TUVX.get_radiation_field_updater()` added to `python/musica/tuvx/tuvx.py`, returning `None` (or raising, whichever fits the wrapper's existing error conventions better) when the configured solver isn't `"from host"`
- An integration test in `python/test/integration/` reproduces tuv-x's own `solver_from_host` unit test: known extraterrestrial flux, solar zenith angle, and Earth-Sun distance, checked against the same hand-computed photolysis rate constant formula the tuv-x test uses
- The test also covers calling `update` with the irradiance arguments omitted (dose rates should come back zero) and calling `run` twice with only one `update` in between (second run should reuse the same field, not error)

**Ideas**
- Do not expose the raw arrays as writable numpy views the way `Radiator.optical_depths` etc. do — always go through `update()`, which should make exactly one C++ call. There is no live "GetXPointer" to read this back from correctly; this field is fully host-owned.
- **Axis order needs an explicit decision, not an assumption.** Confirmed from tuv-x directly: these arrays are `(vertical interface, wavelength bin)`, i.e. height/interface-major — the opposite of how `Radiator`'s existing Python properties document their own arrays (`(num_wavelength_sections, num_height_sections)`, wavelength-major). Getting this backwards would silently transpose the field without an error, the same class of bug found in #912/#1016. Decide and document the axis order explicitly rather than copying the Radiator convention out of habit.
- The host must not fold in the extraterrestrial flux or Earth-Sun distance factor; document this loudly in the docstring, since violating it produces plausible-looking but silently wrong rates.
- Reuse a `configs/tuvx/` fixture built for the Fortran-layer integration test (#1021) rather than inventing a second one.
- Consider whether `get_radiation_field_updater()` should cache its result the way `photolysis_rate_names` etc. do, or return a fresh wrapper each call — check what's simplest given the updater has no meaningful state of its own beyond a pointer back into the live solver.

K20shores:
Closes #1022.

The merged Python bindings work (#1026) covered every acceptance criterion in #1022 except one: "calling run twice with only one update in between (second run should reuse the same field, not error)". The existing test called `update()` before each of its two `run()` calls, never the specific single-update-two-runs case.

Adds that missing test: one `update()` call, two `run()` calls, and confirms the second run reproduces the first's photolysis rates exactly, matching TUV-x's documented behavior of silently reusing the last field set.

## Test plan

- [x] `pytest python/test/integration/test_tuvx_radiation_field_updater.py -v` -- 4 passed
- [x] `pytest python/test/ -q` -- 396 passed, 5 skipped, no regressions
- [x] `codespell` and `autopep8` clean

### DIFF
--- a/python/test/integration/test_tuvx_radiation_field_updater.py
+++ b/python/test/integration/test_tuvx_radiation_field_updater.py
@@ -164,6 +164,34 @@ def test_host_supplied_radiation_field_drives_photolysis_rates():
         np.testing.assert_allclose(actual, expected, rtol=1.0e-8, err_msg=f"dose rate {name}")
 
 
+def test_run_reuses_last_field_without_a_new_update():
+    file = find_config_path("tuvx", "host_radiation_field", "config.json")
+    grid_map = musica.GridMap()
+    profile_map = musica.ProfileMap()
+    radiator_map = musica.RadiatorMap()
+    tuvx = musica.TUVX(grid_map, profile_map, radiator_map, config_path=file)
+
+    updater = tuvx.get_radiation_field_updater()
+    assert updater is not None
+
+    n_interfaces = 5
+    n_bins = 6
+    interfaces, bins = np.meshgrid(np.arange(1, n_interfaces + 1), np.arange(1, n_bins + 1))
+    direct_actinic_flux = 0.2 * interfaces + 0.05 * bins
+    upward_actinic_flux = 0.01 * interfaces + 0.0 * bins
+    downward_actinic_flux = 0.03 * bins + 0.0 * interfaces
+
+    updater.update(direct_actinic_flux, upward_actinic_flux, downward_actinic_flux)
+
+    sza_radians = 42.0 * np.pi / 180.0
+    earth_sun_distance = 0.9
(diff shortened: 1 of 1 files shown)


K20shores:
Not worth adding — closing without merging.

PR #1039, opened 2026-08-25
Correct species interfaces

### LINKED ISSUE #1038: Javascript, Python species missing attributes
In doing https://github.com/NCAR/chemistry-cafe/issues/207, I noticed mech config needed upated docs and corrected attributes (https://github.com/NCAR/MechanismConfiguration/issues/313)

Then when making changes in chem cafe, I noticed we were also missing some of those in the JavaScript/Python interface was missing some of these

**Acceptance criteria**
- javasccript/python species interfaces matches expected mech config keys

**Ideas**

K20shores:
Closes #1038 

- adds absolute tolerance to javascript/python/julia interfaces
- adds tests
- adds density to phase species in javascript

One thing I did was take the density off of species which made the miam work more difficult. We probably need to think about how we actually handle density in micm for phases


javascript/wasm/musica.wasm
boulderdaze, line 1: This binary shouldn't be committed in the first place, no?

K20shores: in this case it should be. This way you can reference a branch of musica and use that wasm. Otherwise we have to do a full release of musica. This makes it easy for development

boulderdaze: I see. The build script mentions that the file is excluded via `gitignore`. I think we can update the comment. 
https://github.com/NCAR/musica/blob/main/javascript/CMakeLists.txt#L13-L14

Do you have any concerns about committing this and `musica.js` in terms of bloating repository history?

K20shores: I don't. I find the convenience of being able to target a branch of musica far outweights keeping around these generated files

K20shores: update the cmake comment

boulderdaze (approved):

Issue #1030, opened 2026-08-21
Scope out surface area density


K20shores:
In cam-sima, surface area density needs to be marshaled to musica. Scope out what that looks like

**Acceptance criteria**
- a full listing of all relevant code locations in cam-sima
- a proposed plan to give data to musica and how to use it
- a recommended unit/integration test for cam-sima

**Ideas**

K20shores:
Relevant PRs

- https://github.com/ESCOMP/CAM/pull/1628

Relevant code locations

In [cam-sima](https://github.com/ESCOMP/CAM-SIMA/tree/development)

- [cime_config/namelist_definition_cam.xml](https://github.com/ESCOMP/CAM-SIMA/blob/3db176f5a4613e1606ab09f4a9d2fb46c45312b9/cime_config/namelist_definition_cam.xml#L444-L488)
  - this namelist has the definition of species that will be included for different aerosols
  - this particular section seems to show that we are defining these BAM species
    - `sulf`, `dust1`, `dust2`, `dust3`, `dust4`, `bcar1`, `bcar2`, `ocar1`, `ocar2`, `sslt1`, `sslt2`, `sslt3`, `sslt4`
    - when a configuration like this is present, we will need to connect them to species in musica and get the properties we need for our [surface reactions](https://mechanismconfiguration.readthedocs.io/en/latest/v1/reactions/surface.html)
    - this means we will need a way to connect names configured in musica configurations with names configured in the nameilst. We might do this similar to how we connect rates in tuvx with rates in micm
- [src/aerosol/radiative_aerosol_definitions.F90](https://github.com/ESCOMP/CAM-SIMA/blob/development/src/aerosol/radiative_aerosol_definitions.F90#L15-L34)
  - This shows that there are only a couple of allowed species names in cam-sima. 
- [src/aerosol/aerosol_state_mod.F90](https://github.com/ESCOMP/CAM-SIMA/blob/development/src/aerosol/aerosol_state_mod.F90)
  - specifically the [surface area density and effective radius attributes](https://github.com/ESCOMP/CAM-SIMA/blob/3db176f5a4613e1606ab09f4a9d2fb46c45312b9/src/aerosol/aerosol_state_mod.F90#L350-L351)
- [src/aerosol/modal_aerosol_state_mod.F90](https://github.com/ESCOMP/CAM-SIMA/blob/development/src/aerosol/modal_aerosol_state_mod.F90)
  - mam implementaiton
[...]

PR #1044, opened 2026-09-09
Update TS1 config file

### LINKED ISSUE #1041: Update TS1 Config file
Our TS1 file is likely old. Update to one downloaded from chem cafe. Use tag MZ391_T1S2_20260129

**Acceptance criteria**
- a v1 version of TS1 from MZ391_T1S2_20260129 is created and all tests pass

**Ideas**
- probably i'll have to add some species or something
- maybe check if any of the 'het rate' rates are in the [ts1 rate constant translatoin](https://drive.google.com/file/d/1xmVFo4_VY2Ro1-8FQKr2KAmxbpapRy6D/view?usp=drive_link)

K20shores:
Closes #1041 

I converted the music box configuration listed in the issue with the musica-cli. To do so successfully, I had to add three species to the species list.

Then, I worked with claude to add in the 'unsupported reactions' section of the configuration file from chemistry cafe. I added those if they could have real values, or with types that let us set their rate to zero so we can figure out how to handle them later. Below lists what each reaction is, how it's handled, and what we might need to do

Correctly implemented, I think:

- usr_N2O5_aer, usr_NO3_aer, usr_NO2_aer, usr_HO2_aer, usr_GLYOXAL_aer, usr_ISOPNITA_aer, usr_ISOPNITB_aer, usr_ONITR_aer, usr_HONITR_aer, usr_TERPNIT_aer, usr_NTERPOOH_aer, usr_NC4CHO_aer, usr_NC4CH2OH_aer
  - Tropospheric aerosol uptake, one gas-phase reactant each
  - Implemented as SURFACE with constant reaction probabilities from CAM 
- het7, het8, het11
  - N2O5, ClONO2, BrONO2 hydrolysis on NAT polar stratospheric cloud particles
  - Implemented as SURFACE with constant reaction probabilities from CAM
- het12, het13, het14
  - Same three reactions, on ice polar stratospheric cloud particles
  - Implemented as SURFACE with constant reaction probabilities from CAM

Deferred for later, but still included

- het1–3
  - The uptake rate depends on temperature and aerosol composition
  - Implementing CAM's composition/temperature-dependent formula would fix this properly.
    - Or, if these can be computed using the surface area density/effective radius provided to use from CAM/CAM-SIMA, we should do that instead
  - Implemented as SURFACE with reaction probability 0
- het4–6
  - MICM's SURFACE type only allows one gas-phase reactant. These need 2.
  - Fixing this needs both the composition-dependent formula above and a new reaction type that can consume two reactants.
  - Implemented as USER_DEFINED with scaling factor 0
- het9, het10
  - Same two-reactant limitation
  - Implemented as USER_DEFINED with scaling factor 0
- het15–17
  - Same situation as het9-10
  - Implemented as USER_DEFINED with scaling factor 0
- [usr_CO_OH](https://github.com/ESCOMP/CAM/blob/ab476f9b7345cbefdc4cf67ff17f0fe85d8c7387/src/chemistry/mozart/mo_usrrxt.F90#L920-L957): CO + OH → CO2 + HO2. JPL's official rate law is two coupled pieces.
  - I guess we used to be able to represent this before [JPL19](https://science.jpl.nasa.gov/documents/1487/NASA-JPL_Evaluation_19-5.pdf) as the sum of two reactions
  - The updated formula in JPL 19 prevents that because it's a Troe term, plus a correction term that subtracts a fraction of that same Troe term
  - no existing MICM reaction type can compute that subtraction.
  - It's USER_DEFINED with scaling factor 0 for now, and needs a new MICM reaction type, or somehow we need to represent this as a mix of several existing types if it can be
  - Implemented as USER_DEFINED with scaling factor 0

AI usage: Claude helped me correctly set the names of photolysis rates and to either implement the het/user define rates or add stubs


jimmielin:
Thanks @K20shores for sending this over.

A few general chemistry notes that stood out to me then specifically for CAM-SIMA.

1. `usr_SO3_H2O`

https://github.com/NCAR/musica/blob/10d504bc0be0e7d44906c75ade365fca92939fd6/configs/v1/ts1/t1s2.json#L3985-L4013

I think the `A` here is mechanistically wrong. It appears that `5.1188e-23` was arrived here by `8.5e-41 * 6.022e17` from CAM

```fortran
!-----------------------------------------------------------------
!       ... so3 + 2*h2o --> h2so4 + h2o
!       Note: this reaction proceeds by the 2 intermediate steps below
!           so3 + h2o --> adduct
!           adduct + h2o --> h2so4 + h2o
!               (Lovejoy et al., JCP, pp. 19911-19916, 1996)
!	The first order rate constant used here is recommended by JPL 2011.
!	This rate involves the water vapor number density.
!-----------------------------------------------------------------

       if ( usr_SO3_H2O_ndx > 0 ) then
          call comp_exp( exp_fac, 6540.0_r8*tinv(:), ncol )
          if( h2o_ndx > 0 ) then
             fc(:) = 8.5e-21_r8 * m(:,k) * h2ovmr(:,k) * exp_fac(:)
          else
             fc(:) = 8.5e-21_r8 * invariants(:,k,inv_h2o_ndx) * exp_fac(:)
          end if
          rxt(:,k,usr_SO3_H2O_ndx) = 1.0e-20_r8 * fc(:)
       end if
```

where `6.022e17` is the conversion factor for cm3 molec-1 s-1 to (mol m-3)-1 s-1 ("`m`") for a **two-body** rxn constant but the actual mechanics of this reaction is a fast pre-equilibrium

SO3 + H2O <-> SO3.H2O (adduct fast pre-equilibrium)
SO3.H2O + H2O -> H2SO4 + H2O

the steady state observed rate is `k(T) [SO3] [H2O]^2`, where `k(T)` is the `fc(:)` here. Note how it depends on `[H2O]^2`, but one of the `[H2O]` factors got entirely dropped in the MUSICA mechanism if you write it out as just `SO3 + H2O -> H2SO4 + H2O` (as in the .json, the "2" also got dropped from the reactants and needs to be put back)... we can't just multiply the constants, leaving out one of the H2Os is off by a factor of `[H2O]`... so about 17 orders of magnitude.

2. `usr_CO_OH` being stubbed out makes the mechanism basically unrealistic as written because CO + OH is the dominant global OH sink (and the only CO oxidation path in TS1, I think?) -- so all the tropospheric oxidants are meaningless until this is fixed.

Either a new MICM type has to be made or fall back to the pre-JPL19 two-term representation. Zero will really mess up the mechanism...

3. the het stubs are fine but this means no strat het chemistry and no ozone-hole chemistry.

More notes on SIMA forthcoming...

jimmielin:
For CAM-SIMA implementation...

Right now the config has one SURFACE reaction per gas, so there's only one pair of (radius, number). This is at odds with what we agreed on for https://github.com/NCAR/musica/issues/1030 where we can't collapse per-mode SAD into one aggregate, so if you're not modifying MICM internals, per-mode duplicates of the same reactions needs to be made to match CAM.

Thinking more about the het.. reactions I realize that the het1-3 being deferred might be problematic, since N2O5 hydrolysis is a dominant pathway to produce HNO3 globally. 

I think het1-17 would be better wired up as `USER_DEFINED`. For het1-3, have the host supply `k` which is a function of SAD, so you don't have to have variable gamma for `SURFACE` reactions. For het4 it's trickier as there's a limiting reagent construction

```fortran
!-----------------------------------------------------------------------
!     	... ClONO2 + HCl(liq) =  Cl2  + HNO3  Sulfate Aerosol Reaction
!-----------------------------------------------------------------------
                  if( hclvmr > small_div .and. clono2vmr > small_div ) then
                     if ( hclvmr > clono2vmr ) then
                        rxt(i,k,rid_het4) = max( 0._r8,wrk*av_clono2*gprob_cnt_hcl(i,k) )*hcldeni
                     else
                        rxt(i,k,rid_het4) = max( 0._r8,wrk*av_clono2*gprob_cnt_hcl(i,k) )*cntdeni
                     end if
                  end if
```

For het13, 14 I see it's implemented now but my hunch is that MICM's rates may be too slow for ice PSCs. MICM's [SURFACE formula is the full resistor](https://mechanismconfiguration.readthedocs.io/en/latest/v1/reactions/surface.html) but for CAM stratospheric het13, 14, it is the free-molecular construction with no gas-diffusion term. This differs from the trop rates in TS1 (e.g., usr_N2O5_aer...)

K20shores:
> Thanks @K20shores for sending this over.
> 
> A few general chemistry notes that stood out to me then specifically for CAM-SIMA.
> 
> 1. `usr_SO3_H2O`
> 
> https://github.com/NCAR/musica/blob/10d504bc0be0e7d44906c75ade365fca92939fd6/configs/v1/ts1/t1s2.json#L3985-L4013
> 
> I think the `A` here is mechanistically wrong. It appears that `5.1188e-23` was arrived here by `8.5e-41 * 6.022e17` from CAM
> 
> ```fortran
> !-----------------------------------------------------------------
> !       ... so3 + 2*h2o --> h2so4 + h2o
> !       Note: this reaction proceeds by the 2 intermediate steps below
> !           so3 + h2o --> adduct
> !           adduct + h2o --> h2so4 + h2o
> !               (Lovejoy et al., JCP, pp. 19911-19916, 1996)
> !	The first order rate constant used here is recommended by JPL 2011.
> !	This rate involves the water vapor number density.
> !-----------------------------------------------------------------
> 
>        if ( usr_SO3_H2O_ndx > 0 ) then
>           call comp_exp( exp_fac, 6540.0_r8*tinv(:), ncol )
>           if( h2o_ndx > 0 ) then
>              fc(:) = 8.5e-21_r8 * m(:,k) * h2ovmr(:,k) * exp_fac(:)
>           else
>              fc(:) = 8.5e-21_r8 * invariants(:,k,inv_h2o_ndx) * exp_fac(:)
>           end if
>           rxt(:,k,usr_SO3_H2O_ndx) = 1.0e-20_r8 * fc(:)
>        end if
> ```
> 
> where `6.022e17` is the conversion factor for cm3 molec-1 s-1 to (mol m-3)-1 s-1 ("`m`") for a **two-body** rxn constant but the actual mechanics of this reaction is a fast pre-equilibrium
> 
> SO3 + H2O <-> SO3.H2O (adduct fast pre-equilibrium) SO3.H2O + H2O -> H2SO4 + H2O
> 
> the steady state observed rate is `k(T) [SO3] [H2O]^2`, where `k(T)` is the `fc(:)` here. Note how it depends on `[H2O]^2`, but one of the `[H2O]` factors got entirely dropped in the MUSICA mechanism if you write it out as just `SO3 + H2O -> H2SO4 + H2O` (as in the .json, the "2" also got dropped from the reactants and needs to be put back)... we can't just multiply the constants, leaving out one of the H2Os is off by a factor of `[H2O]`... so about 17 orders of magnitude.
> 
> 2. `usr_CO_OH` being stubbed out makes the mechanism basically unrealistic as written because CO + OH is the dominant global OH sink (and the only CO oxidation path in TS1, I think?) -- so all the tropospheric oxidants are meaningless until this is fixed.
> 
> Either a new MICM type has to be made or fall back to the pre-JPL19 two-term representation. Zero will really mess up the mechanism...
> 
> 3. the het stubs are fine but this means no strat het chemistry and no ozone-hole chemistry.
> 
> More notes on SIMA forthcoming...

Ah, looks like the coefficient was dropped. Not sure how that happened, but fixing that bug will correct this rate

jimmielin:
> Ah, looks like the coefficient was dropped. Not sure how that happened, but fixing that bug will correct this rate

`A` also has to be rescaled to 8.5e-41 * m^2. If you're taking `m ~ 6.022e17` then `A` would be around `3e-5`... Claude can probably work out the right coeff. based on the CAM code.

K20shores:
> > Ah, looks like the coefficient was dropped. Not sure how that happened, but fixing that bug will correct this rate
> 
> `A` also has to be rescaled to 8.5e-41 * m^2. If you're taking `m ~ 6.022e17` then `A` would be around `3e-5`... Claude can probably work out the right coeff. based on the CAM code.

I don't think we need Claude for this. This conversion is done using avogadro's number and I'm pretty sure it's just a bug somewhere in mechanism configuration that shouldn't be too hard to find.

K20shores:
> > > Ah, looks like the coefficient was dropped. Not sure how that happened, but fixing that bug will correct this rate
> > 
> > 
> > `A` also has to be rescaled to 8.5e-41 * m^2. If you're taking `m ~ 6.022e17` then `A` would be around `3e-5`... Claude can probably work out the right coeff. based on the CAM code.
> 
> I don't think we need Claude for this. This conversion is done using avogadro's number and I'm pretty sure it's just a bug somewhere in mechanism configuration that shouldn't be too hard to find.

or I'm misunderstanding what you're saying entirely

K20shores:
> Thanks @K20shores for sending this over.
> 
> A few general chemistry notes that stood out to me then specifically for CAM-SIMA.
> 
> 1. `usr_SO3_H2O`
> 
> https://github.com/NCAR/musica/blob/10d504bc0be0e7d44906c75ade365fca92939fd6/configs/v1/ts1/t1s2.json#L3985-L4013
> 
> I think the `A` here is mechanistically wrong. It appears that `5.1188e-23` was arrived here by `8.5e-41 * 6.022e17` from CAM
> 
> ```fortran
> !-----------------------------------------------------------------
> !       ... so3 + 2*h2o --> h2so4 + h2o
> !       Note: this reaction proceeds by the 2 intermediate steps below
> !           so3 + h2o --> adduct
> !           adduct + h2o --> h2so4 + h2o
> !               (Lovejoy et al., JCP, pp. 19911-19916, 1996)
> !	The first order rate constant used here is recommended by JPL 2011.
> !	This rate involves the water vapor number density.
> !-----------------------------------------------------------------
> 
>        if ( usr_SO3_H2O_ndx > 0 ) then
>           call comp_exp( exp_fac, 6540.0_r8*tinv(:), ncol )
>           if( h2o_ndx > 0 ) then
>              fc(:) = 8.5e-21_r8 * m(:,k) * h2ovmr(:,k) * exp_fac(:)
>           else
>              fc(:) = 8.5e-21_r8 * invariants(:,k,inv_h2o_ndx) * exp_fac(:)
>           end if
>           rxt(:,k,usr_SO3_H2O_ndx) = 1.0e-20_r8 * fc(:)
>        end if
> ```
> 
> where `6.022e17` is the conversion factor for cm3 molec-1 s-1 to (mol m-3)-1 s-1 ("`m`") for a **two-body** rxn constant but the actual mechanics of this reaction is a fast pre-equilibrium
> 
> SO3 + H2O <-> SO3.H2O (adduct fast pre-equilibrium) SO3.H2O + H2O -> H2SO4 + H2O
> 
> the steady state observed rate is `k(T) [SO3] [H2O]^2`, where `k(T)` is the `fc(:)` here. Note how it depends on `[H2O]^2`, but one of the `[H2O]` factors got entirely dropped in the MUSICA mechanism if you write it out as just `SO3 + H2O -> H2SO4 + H2O` (as in the .json, the "2" also got dropped from the reactants and needs to be put back)... we can't just multiply the constants, leaving out one of the H2Os is off by a factor of `[H2O]`... so about 17 orders of magnitude.
> 
> 2. `usr_CO_OH` being stubbed out makes the mechanism basically unrealistic as written because CO + OH is the dominant global OH sink (and the only CO oxidation path in TS1, I think?) -- so all the tropospheric oxidants are meaningless until this is fixed.
> 
> Either a new MICM type has to be made or fall back to the pre-JPL19 two-term representation. Zero will really mess up the mechanism...
> 
> 3. the het stubs are fine but this means no strat het chemistry and no ozone-hole chemistry.
> 
> More notes on SIMA forthcoming...

Ah, I found it. Not a bug in musica, but the old chemistry cafe. It wrote out this reaction `H2O + SO3 -> H2SO4 + H2O`, so it was missing the stoichiometric coefficient of 2. When I correct it to be `2H2O + SO3 -> H2SO4 + H2O` and run our conversion tool again, we get this

```json
{
            "type": "ARRHENIUS",
            "A": 3.082625243326537e-05,
            "B": 0.0,
            "C": 6539.99676963515,
            "D": 300.0,
            "E": 0.0,
            "reactants": [
                {
                    "name": "H2O",
                    "coefficient": 2.0
                },
                {
                    "name": "SO3",
                    "coefficient": 1.0
                }
            ],
            "products": [
                {
                    "name": "H2SO4",
                    "coefficient": 1.0
                },
                {
                    "name": "H2O",
                    "coefficient": 1.0
                }
            ],
            "gas phase": "gas"
        }
```

which seems correct, i think

jimmielin:
> When I correct it to be 2H2O + SO3 -> H2SO4 + H2O and run our conversion tool again, 

Ah, neat, since it gets to `"A": 3.082625243326537e-05,` the same as my back of envelope calculation I think that's correct. I didn't know you generated the coeffs programmatically too, that's really neat. If I did it by hand I would probably asked Claude to confirm :-)

K20shores:
> > When I correct it to be 2H2O + SO3 -> H2SO4 + H2O and run our conversion tool again,
> 
> Ah, neat, since it gets to `"A": 3.082625243326537e-05,` the same as my back of envelope calculation I think that's correct. I didn't know you generated the coeffs programmatically too, that's really neat. If I did it by hand I would probably asked Claude to confirm :-)

Our conversion tool reads what Chemistry Cafe outputs with our mechanism configuration library. It converts from the old camp format to the newer format that uses SI units everywhere. It applies those conversions automatically. Louisa creates this information in Chemistry Cafe. This is likely a case where it's correct in Chemistry Cafe for CAM and potentially incorrect for the other output targets of Chemistry Cafe (kpp, box mox, music box, something else) because the main target is CAM, so it relies on Fortran code being correct in CAM rather than relying entirely on data, as we do for the mechanism in musica.

boulderdaze (approved):

boulderdaze (approved):

K20shores:
We will need to add a new reaction rate to support the newest version. See https://github.com/NCAR/micm/issues/1092. This will need to happen in mechanism configuration and musica, and I guess chemistry cafe and music box interactive

